PR #121, opened 2022-02-11
Update specimen function

anthonypetersen:
* modified specimenExists() in firestorm.js to not update document if a match is found
* added "updateSpecimen" case to biospecimen.js


bhaumik55231 (approved):
Looks great!

PR #134, opened 2022-03-23
Shipping Dashboard: Added UpdateBox API & Updated AddBox API

abhinavjonnada82:
This PR addresses following feature:
-> Added UpdateBox API & Updated AddBox API endpoints
-> Edited BoxExists function to return only True or False

Checklist:
- [x] Code cleanup
- [x] Check for ES Lint warnings
- [ ] Verify test cases
- [x] Check for GIT conflicts
- [ ] Verified unit tests pass with current changes
- [ ] Any dependencies or modules added
- [ ] Attach PoC
- [x] Notes added

**Notes:**

_**http://{URL}/biospecimen?api=addBox**_ returns success if box is added
_**http://{URL}/biospecimen?api=updateBox**_ returns success if existing box is updated

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -198,8 +198,8 @@ const biospecimenAPIs = async (req, res) => {
             const exists = await boxExists(boxId, loginSite, requestData);
             if (exists === true) return res.status(400).json(getResponseJSON('Box already exists!', 400));
             if (exists === false) {
-                const { storeBox } = require('./firestore');
-                await storeBox(requestData);
+                const { addBox } = require('./firestore');
+                await addBox(requestData);
             }
         }
         return res.status(200).json({message: 'Addded!', code:200})
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -189,13 +189,13 @@ const biospecimenAPIs = async (req, res) => {
         }
         const requestData = req.body;
         if(Object.keys(requestData).length === 0 ) return res.status(400).json(getResponseJSON('Request body is empty!', 400));
-        
+        console.log('1222')
         if(requestData['132929440']){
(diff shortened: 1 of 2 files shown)


utils/biospecimen.js
```diff
@@ -191,20 +191,40 @@ const biospecimenAPIs = async (req, res) => {
         }
         const requestData = req.body;
         if(Object.keys(requestData).length === 0 ) return res.status(400).json(getResponseJSON('Request body is empty!', 400));
+        if(requestData['132929440']){
+            const boxId = requestData['132929440'];
+            const loginSite = requestData['789843387']
+            const { boxExists } = require('./firestore');
+            const exists = await boxExists(boxId, loginSite, requestData);
```
anthonypetersen, line 198: probably being super picky, but not sure we need to pass requestData into "boxExists"

anthonypetersen (changes requested):
made a few comments regarding parameters in the calling matching what we have in the function declarations

utils/biospecimen.js
```diff
@@ -191,20 +191,40 @@ const biospecimenAPIs = async (req, res) => {
...
+            if (exists === true) return res.status(400).json(getResponseJSON('Box already exists!', 400));
+            if (exists === false) {
+                const { addBox } = require('./firestore');
+                await addBox(requestData);
+            }
+        }
+        return res.status(200).json({message: 'Addded!', code:200})
```
Gbarra9, line 205: I think there is a typo for the message 'Addded!'
 
Does the naming convention  for a status code of 200 have to be "Success!" or can it be anything that indicates a successful request?

abhinavjonnada82: AFAIK its a message so can be anything. Anyways I changed it back to Success!

PR #162, opened 2022-05-16
Modify getBoxesPagination and getNumBoxesShipped for biospecimen shipping reports query page

Gbarra9:
The pull request addresses the following:
- Modified both getBoxesPagination and getNumBoxesShipped functions
- Changed queries made to firestore by changing the siteAcronym to siteCode
- Changed startDate and endDate variables (if no endDate and startDate is present in POST body) to a default number with string data type
- startDate and endDate default values changed to compare "<=" or ">=" with date [ISO format(ISO 8601)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -362,7 +362,7 @@ const biospecimenAPIs = async (req, res) => {
         }
         const {getBoxesPagination} = require('./firestore');
         const requestData = req.body;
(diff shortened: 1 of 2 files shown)


abhinavjonnada82 (approved):

we-ai (approved):
Looks good to update siteAcronym to siteCode, and use strings for startDate and endDate.

PR #219, opened 2022-07-05
Stage / Prod Sync 070722

anthonypetersen:


anthonypetersen:
Changes included in this release:

* added "getUserProfile" API endpoint for Biospecimen Dashboard
* added logic in retrieveParticipantsByStatus() to handle different operators and value types
* bug fix in setPackageReceiptFedex()
* added logic in notificationHandler() to evaluate Specification Primary Field if it's in ISO date / time format
* moved checkDefaultFlags() function from SM Dashboard to shared.js
* added logic to submit() to delete fields if the value passed in is undefined
* updated PWA "getUserProfile" endpoint to utilized checkDefaultFlags() function

PR #303, opened 2022-11-17
changes for BQ fix

anthonypetersen:
connectApp

		* added endpoint getUserSurveys

submission

		* added function getUserSurveys
		* added function submitSurvey
		* added logic to call checkDerivedVariables() after survey completion
		* added separate path for submit() if we are submitting survey answer data

firestore

		* added function retrieveConnectID
		* added function surveyExists
		* added function storeSurvey
		* added function updateSurvey
		* added function retrieveUserSurveys

shared

		* added constant moduleConceptsToCollections
		* added constant moduleStatusConcepts
		* added function cleanSurveyData

validation

		* updated checkDerivedVariables to calculate derivation for enabling menstrual cycle survey
		* added logic to not update participant if there isn't anything to update

### DIFF
--- a/utils/shared.js
+++ b/utils/shared.js
@@ -248,15 +248,13 @@ const moduleConceptsToCollections = {
 }
 
 const moduleStatusConcepts = {
-    "949302066" :       "module1_v1",
-    "949302066" :       "module1_v2",
-    "536735468" :       "module2_v1",
-    "536735468" :       "module2_v2",
-    "976570371" :       "module3_v1",
-    "663265240" :       "module4_v1",
-    "265193023" :       "bioSurvey_v1",
-    "459098666" :       "menstrualSurvey_v1",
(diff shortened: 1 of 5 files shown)


utils/shared.js
```diff
@@ -247,6 +247,18 @@ const moduleConceptsToCollections = {
     "D_826163434" :     "clinicalBioSurvey_v1"
 }
 
+const moduleStatusConcepts = {
+    "949302066" :       "module1_v1",
```
we-ai, line 251: duplicate keys

utils/shared.js
```diff
@@ -247,6 +247,18 @@ const moduleConceptsToCollections = {
...
+    "949302066" :       "module1_v2",
```
we-ai, line 252: duplicate keys

PR #316, opened 2023-01-04
Clinical Dashboard: Enhanced error handling accessionIdExists api

abhinavjonnada82:
Title^^

When accession ID doesn't exist returns error code 400 and in turn terminates the process on the clinical dashboard


anthonypetersen (commented):
This is fine, but I would prefer the dashboard side of this to not be reading what the message is. Instead check to see if the returning payload contains a "data" header.

PR #322, opened 2023-02-01
Building Consistency Check Logic into submitParticipantsData API

anthonypetersen:
This PR is to fix issue https://github.com/episphere/connect/issues/607

* Added file `submitParticipantData.json` to define "rules" for using Submit Participant Data API
* Refactored function `submitParticipantsData` to more closely resemble structure of `updateParticipantData`
* Added logic in `submitParticipantsData` to utilize new JSON Rules file to add Consistency Checking into API


utils/sites.js
```diff
@@ -24,60 +23,121 @@ const submitParticipantsData = async (req, res, site) => {
...
         }
+
         siteCode = authorized.siteCode;
     }
 
-    if(req.body.data === undefined) return res.status(400).json(getResponseJSON('data is undefined in request body.', 400));
-    if(req.body.data.length === undefined || req.body.data.length < 1) return res.status(400).json(getResponseJSON('data array doesn\'t have any element.', 400));
+    if(req.body.data === undefined) return res.status(400).json(getResponseJSON('Bad request. Data is not defined in request body.', 400));
+    if(!Array.isArray(req.body.data)) return res.status(400).json(getResponseJSON('Bad request. Data must be an array.', 400));
+    if(req.body.data.length === undefined || req.body.data.length < 1) return res.status(400).json(getResponseJSON('Bad request. Data array does not have any elements.', 400));
```
we-ai, line 32: `if (req.body.data.length === 0)`

utils/sites.js
```diff
@@ -24,60 +23,121 @@ const submitParticipantsData = async (req, res, site) => {
...
+    for(let dataObj of dataArray){
+        if(dataObj.token === undefined) {
+            error = true;
+            responseArray.push({'Invalid Request': {'Token': 'UNDEFINED', 'Errors': 'Token not defined in data object.'}});
+            continue;
+        }
+        
+        const participantToken = dataObj.token;
+        const { getParticipantData } = require('./firestore');
+        const record = await getParticipantData(participantToken, siteCode);
```
we-ai, line 53: Should we get data of participants in one query?

utils/sites.js
```diff
@@ -24,60 +23,121 @@ const submitParticipantsData = async (req, res, site) => {
...
+
+        const docID = record.id;
+        const docData = record.data;
+
+        const flat = (obj, att, attribute) => {
+            for(let k in obj) {
+                if(typeof(obj[k]) === 'object') flat(obj[k], att, attribute ? `${attribute}.${k}`: k)
+                else flattened[att][attribute ? `${attribute}.${k}`: k] = obj[k]
+            }
+        }
```
we-ai, line 69: The function can be moved out of the loop, so that it doesn't need to be re-defined in each iteration.

utils/sites.js
```diff
@@ -24,60 +23,121 @@ const submitParticipantsData = async (req, res, site) => {
...
                     }
-                    // Make participant active if Study invitaion is sent.
-                    if(key === '934298480' && record.data['512820379'] !== 854703046) { // If age deidentified data is provided and participant is not passive then make this participant Active
-                        newStateElements['512820379'] = 486306141;
-                        newStateElements['471593703'] = new Date().toISOString();
+                    else {
+                        if(submitRules[key].dataType !== typeof dataObj[key]) {
+                            errors.push(" Invalid data type for Key (" + key + ")");
+                        }
+                        else {
```
we-ai, line 99: `else if`

utils/sites.js
```diff
@@ -24,60 +23,121 @@ const submitParticipantsData = async (req, res, site) => {
...
+                            if(submitRules[key].values) {
+                                if(submitRules[key].values.filter(value => value.toString() === dataObj[key].toString()).length == 0) {
```
we-ai, line 101: `if (!submitRules[key].values.includes(dataObj[key]))`

utils/sites.js
anthonypetersen: would this capture the scenario that length is undefined? or are you just suggesting I replace "req.body.data.length < 1" with "req.body.data.length === 0" ?

utils/sites.js
anthonypetersen: the way you build / run queries in Firestore I'm not sure it would make a difference.. Firestore doesn't support OR conditionality, and I believe if you used "IN" it would be the same as running an individual query for each one.. I may be mistaken though.

utils/sites.js
anthonypetersen: Noted

utils/sites.js
anthonypetersen: I don't believe this need to be an "ELSE IF", we want everything else flowing to this path if not captured in the IF statement on line 96

utils/sites.js
anthonypetersen: I will try this out, thanks.

utils/sites.js
anthonypetersen: Great point, thank you for pointing this out 👍

utils/sites.js
we-ai: Nothing wrong here. I just put my personal preference.

PR #341, opened 2023-04-11
SiteManager Dashboard: Added API endpoint for switch login mechanism

abhinavjonnada82:
This PR addresses the following features:
-> Added API endpoint to switch login method from email to phone & vice-versa
-> Updated existing login method (email or phone)

Checklist:
- [x] Code cleanup
- [x] Check for ES Lint warnings
- [x] Verify test cases
- [x] Check for GIT conflicts
- [ ] Verified unit tests pass with current changes
- [x] Any dependencies or modules added
- [ ] Attach PoC
- [x] Notes added

**Notes:**
_**http://{URL}/dashboard?api=updateSignMechanism**_

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1866,11 +1866,14 @@ const verifyUsersEmailOrPhone = async (req) => {
             return false;
         }
     }
+}
 
-    const updateUsersCurrentEmail = async (email, uid) => {    
+const updateUsersCurrentLogin = async (req, uid) => {   
+    const queries = req
+    if (queries.email) {
         try {
             await admin.auth().updateUser(uid,
-                {       email: email, 
+                {       email: queries.email,
                 })
             return true
         }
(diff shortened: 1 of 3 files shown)


utils/sites.js
```diff
@@ -344,7 +344,35 @@ const qc = (newData, existingData, rules) => {
     return errors;
 }
 
+const updateSignMechanism = async (req, res) => {
```
anthonypetersen, line 347: Not to be super picky, but please update this.. I know what you're saying, but "sign mechanism" might be initially confusing to someone down the road. Suggestions at the top of my head would be something like "updateUserFirebaseAuthentication".

utils/sites.js
```diff
@@ -152,7 +152,7 @@ const submitParticipantsData = async (req, res, site) => {
         responseArray.push({'Success': {'Token': participantToken, 'Errors': 'None'}});
     }
 
-    return res.status(error ? 206 : 200).json({code: error ? 206 : 200, results: responseArray});
+    return res.status(error ? 206 : 200).json({code: error ? 206 : 200, results: responseArray}); // 206 one or more pt error
```
anthonypetersen, line 155: is this comment needed or was this leftover from when you were doing the API documentation?

utils/sites.js
```diff
@@ -344,7 +344,35 @@ const qc = (newData, existingData, rules) => {
...
+const updateSignMechanism = async (req, res) => {
+    logIPAdddress(req);
+    setHeaders(res);
+    
+    if(req.method === 'OPTIONS') return res.status(200).json({code: 200});
+
+    if(req.method !== 'POST') {
+        return res.status(405).json(getResponseJSON('Only POST requests are accepted!', 405));
+    }
+        const { APIAuthorization } = require('./shared');
```
anthonypetersen, line 356: un-tab all of these lines to make them even with code above it

utils/sites.js
```diff
@@ -344,7 +344,35 @@ const qc = (newData, existingData, rules) => {
...
+        const authorized = await APIAuthorization(req);
+        if(authorized instanceof Error) return res.status(500).json(getResponseJSON(authorized.message, 500));
+        if(!authorized) return res.status(401).json(getResponseJSON('Authorization failed!', 401));
+        if(req.body.data === undefined) return res.status(400).json(getResponseJSON('Bad request. Data is not defined in request body.', 400));
+        const { updateUserPhoneSigninMethod, updateUserEmailSigninMethod, updateUsersCurrentEmail, updateUsersCurrentPhone } = require('./firestore');
+        let status = ``
+        if (req.body.data.hasOwnProperty('phone') && req.body.data.flag === `replaceSignin`) status = await updateUserPhoneSigninMethod(req.body.data.phone, req.body.data.uid);
```
anthonypetersen, line 363: Double check with @we-ai ... I think he provided an argument previously against using "hasOwnProperty()"

utils/firestore.js
```diff
@@ -1866,7 +1866,73 @@ const verifyUsersEmailOrPhone = async (req) => {
             return false;
         }
     }
-}
+
+    const updateUsersCurrentEmail = async (email, uid) => {
```
anthonypetersen, line 1870: I'm curious what @we-ai thinks as well, but with how similar these four functions are I would rather see these combined into one function. This would also make the code that calls this function "simpler".

anthonypetersen (changes requested):

anthonypetersen:
Also, because this is going to be triggered from the SM Dashboard the entry point should be from dashboard.js

utils/sites.js
we-ai: I agree `updateSignMechanism` is not clear enough. I'd prefer `updateSignInMethod`, `updateUserAuthentication`, etc.

abhinavjonnada82:
> Also, because this is going to be triggered from the SM Dashboard the entry point should be from dashboard.js

Done

utils/firestore.js
abhinavjonnada82: I combined Update current email & phone functions together. I prefer leaving replace phone & email separate. As the parameters to replace email & phone are different from each other.

anthonypetersen (approved):
looks good, thanks for the updates

PR #352, opened 2023-05-10
Enabling Array Payloads - Update Participant Data API

anthonypetersen:
This PR resolves issue https://github.com/episphere/connect/issues/639

* Add new keys to `updateParticipantData.json`
* Add rules to `qc()` to allow value types of `array` to be accepted
* Arrays must contain only string objects
* Preventing flattening of keys with values of type `array` when flattening new data

### DIFF
--- a/utils/sites.js
+++ b/utils/sites.js
@@ -245,7 +245,7 @@ const updateParticipantData = async (req, res, authObj) => {
         
             if(primaryIdentifiers.indexOf(key) !== -1) continue;
 
-            if(typeof(dataObj[key]) === 'object') flat(dataObj[key], 'newData', key);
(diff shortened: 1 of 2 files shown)


utils/sites.js
```diff
@@ -323,6 +323,19 @@ const qc = (newData, existingData, rules) => {
                         errors.push(" Invalid data type / format for Key (" + key + ")");
                     }
                 }
+                else if(rules[key].dataType == 'array') {
```
we-ai, line 326:
```suggestion
                else if(rules[key].dataType === 'array') {
```

we-ai (approved):
Looks good.

PR #361, opened 2023-05-23
Update connectFaas for connect app user editing

JoeArmani:
This PR adds an endpoint for logged-in connectApp users to update their firebase Authentication method in connectApp.

This is in continued support of [issue #635](https://github.com/episphere/connect/issues/635)

Supported login methods:
-email
-phone

Changes:
(1) Add reference to `updateUserFirebaseAuthentication` in `connectApp.js`
(2) Add `updateUserFirebaseAuthentication()` to `firestore.js`

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1856,15 +1856,27 @@ const updateParticipantFirebaseAuthentication = async (req, res) => {
     if(req.method !== 'POST') {
         return res.status(405).json(getResponseJSON('Only POST requests are accepted!', 405));
     }
+    const data = req.body.data;
+    const flag = data.flag;
+    const uid =  data.uid;
 
-    if(req.body.data === undefined) {
+    if(data === undefined) {
         return res.status(400).json(getResponseJSON('Bad request. Data is not defined in request body.', 400));
     }
 
     let status = '';
-    if (req.body.data['phone'] && req.body.data.flag === `replaceSignin`) status = await updateUserPhoneSigninMethod(req.body.data.phone, req.body.data.uid);
-    if (req.body.data['email'] && req.body.data.flag === `replaceSignin`) status = await updateUserEmailSigninMethod(req.body.data.email, req.body.data.uid);
-    if (req.body.data.flag === `updateEmail` || req.body.data.flag === `updatePhone`) status = await updateUsersCurrentLogin(req.body.data, req.body.data.uid);
+    if (flag === `replaceSignin`) {
+        if (data['phone']) {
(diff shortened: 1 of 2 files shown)


Gbarra9 (commented):
Looks good. 

Is the use of the back ticks for assigning an empty string to the status variable a coding preference? 
https://github.com/episphere/connectFaas/pull/361/files#diff-be91e66430aa4dd2f61668d1bd17d3cb449f57947e13bfa2f6bc087919967282R1864

JoeArmani:
> Looks good.
> 
> Is the use of the back ticks for assigning an empty string to the status variable a coding preference? https://github.com/episphere/connectFaas/pull/361/files#diff-be91e66430aa4dd2f61668d1bd17d3cb449f57947e13bfa2f6bc087919967282R1864

Thanks, Gene, I actually meant to update that - just did. I prefer single quotes when I don't need the backticks.

utils/firestore.js
```diff
@@ -1852,6 +1852,27 @@ const verifyUsersEmailOrPhone = async (req) => {
...
+    }
+
+    if(req.body.data === undefined) {
+        return res.status(400).json(getResponseJSON('Bad request. Data is not defined in request body.', 400));
+    }
+
+    let status = '';
+    if (req.body.data['phone'] && req.body.data.flag === `replaceSignin`) status = await updateUserPhoneSigninMethod(req.body.data.phone, req.body.data.uid);
+    if (req.body.data['email'] && req.body.data.flag === `replaceSignin`) status = await updateUserEmailSigninMethod(req.body.data.email, req.body.data.uid);
+    if (req.body.data.flag === `updateEmail` || req.body.data.flag === `updatePhone`) status = await updateUsersCurrentLogin(req.body.data, req.body.data.uid);
```
we-ai, line 1867: Maybe put `else` so that 2rd and 3rd conditions won't be check if the first is true.

we-ai (approved):
Looks good. Put a minor comment.

JoeArmani:
Updated per Warren's comment, merging now

we-ai:
Comment about a small issue after further check.

A call to `updateParticipantFirebaseAuthentication` in 'firestore.js' will break, because `getResponseJSON` is not imported. A quick fix is to import `getResponseJSON`.

A better solution will be moving `updateParticipantFirebaseAuthentication` to a file that handles http requests (eg 'validation.js'), and let functions in 'firebase.js' handle only Firestore database.

JoeArmani:
Continuing update with new PR here: https://github.com/episphere/connectFaas/pull/372

PR #362, opened 2023-05-25
#636 Add function to delete uninvited participants

jhflorey:
The following Pull request (PR) addresses [Automate Deleting Uninvited Participants #636](https://github.com/episphere/connect/issues/636)

Updates to the code:

- Created `removeUninvitedParticipants` function in `utils/firestore.js` to get all participants who are not invited and remove their Firestore document
- Created `participantDataCleanup` file for calling `removeParticipantsDataDestruction` and `removeUninvitedParticipants`
- Created an event `participantDataCleanup` in `index.js`.js
- Changed the expired days of `date data destroy requested` from 60 to 2
- Edited `cloudbuild5.yaml` in `config/dev|stage|prod` for new cloud scheduler topic


JoeArmani (commented):
Update - Comment Edited: formatting, clarity
Question in `firestore.js`: is the batch operation created and committed inside the loop to avoid using a counter? I know there's a 500 doc limit on batch writes.

Can the `batch()` be created once before the loop, updated each iteration of the for loop, then `batch.commit()` once after the loop?

jhflorey:
> Update - Comment Edited: formatting, clarity Question in `firestore.js`: is the batch operation created and committed inside the loop to avoid using a counter? I know there's a 500 doc limit on batch writes.
> 
> Can the `batch()` be created once before the loop, updated each iteration of the for loop, then `batch.commit()` once after the loop?
@JoeArmani  you are correct. i tried to put `batch` before the lood and i ran into 500 doc limit error.

JoeArmani:
> > Update - Comment Edited: formatting, clarity Question in `firestore.js`: is the batch operation created and committed inside the loop to avoid using a counter? I know there's a 500 doc limit on batch writes.
> > Can the `batch()` be created once before the loop, updated each iteration of the for loop, then `batch.commit()` once after the loop?
> > @JoeArmani  you are correct. i tried to put `batch` before the lood and i ran into 500 doc limit error.

Got it. So in that case you would pull limit=500 docs, run the batch, reset the counter, repeat. I wasn't sure how often this process is running. Thanks.

utils/firestore.js
```diff
@@ -453,6 +453,38 @@ const removeParticipantsDataDestruction = async () => {
     }
 }
 
+const removeUninvitedParticipants = async () => {
+    try {
+
+        const currSnapshot = await db
+            .collection('participants')
+            .where('288972510', '==', 353358909)
```
Gbarra9, line 461: I think we can replace the concept Ids for `288972510` and  `353358909` with a reference to [fieldToConceptIdMapping.js](https://github.com/episphere/connectFaas/blob/dev/utils/fieldToConceptIdMapping.js)

jhflorey: Solved

utils/firestore.js
```diff
@@ -453,6 +453,38 @@ const removeParticipantsDataDestruction = async () => {
     }
 }
 
+const removeUninvitedParticipants = async () => {
+    try {
+
+        const currSnapshot = await db
+            .collection('participants')
+            .where(fieldMapping.participantMap.uninvitedRecruits.toString(), '==', fieldMapping.yes)
```
Gbarra9, line 461: This just my personal preference but I like destructuring the fieldMapping references or another variable if they get too big. The only problem with this is that is adds more lines. Maybe we can assign `fieldMapping.participantMap.uninvitedRecruits.toString()` to another variable. 

`const uninvitedRecruits = fieldMapping.participantMap.uninvitedRecruits.toString()`

or 

`const { uninvitedRecruits } = fieldMapping.participantMap`

`.where(uninvitedRecruits.toString() , '==', fieldMapping.yes )`

jhflorey: Solved

Gbarra9 (commented):
Looks good

JoeArmani:
Hi Jessica, this is looking good. I might consider this type of approach just as an optimization of what it's already doing. I think this way we can fetch and process as many documents as needed with one API call. I haven't tested it, but I hope the concepts can contribute.

Limit to 500 docs in the `get()` request. Then I think you can use `startAfter()` to get the next 500 docs. Then, if we use `let` for snapshot we can reassign it after `batch.commit()` in the while loop. NIT: test `===` instead of `==` here - I think `===` is faster but I know `==` covers the case of `fieldMapping.yes` being a string, so what you have is more robust.

```javascript
const removeUninvitedParticipants = async () => {
    try {
        const uninvitedRecruitsCId = fieldMapping.participantMap.uninvitedRecruits.toString();
        const limit = 500;      
        //get the first 500 documents
        let snapshot = await db
            .collection('participants')
            .where(uninvitedRecruitsCId, '==', fieldMapping.yes)
            .limit(limit)
            .get();
        // batch process and continue fetching as long as there are documents
        while (!snapshot.empty) {
            const batch = db.batch();
            snapshot.docs.forEach(doc => {
                const participant = doc.data();
                const fieldKeys = Object.keys(participant);
                const participantRef = doc.ref;
                fieldKeys.forEach(key => {
                    batch.update(participantRef, { [key]: admin.firestore.FieldValue.delete() });
                });
            });

            await batch.commit().catch(err => {
                console.error(`Error occurred when updating documents: ${err}`);
                throw new Error(err);
            });

            // Check whether the snapshot has the limit of 500 docs.
            // If yes, we should check for more docs. 
            // If no, we can break out of the loop because there are no more docs to fetch.
            if (snapshot.size === limit) {
                // mark the last document in the current batch as a placeholder for startAfter() in the next query 
                const lastDocument = snapshot.docs[snapshot.docs.length - 1];

                // Query for the next batch
                snapshot = await db
                    .collection('participants')
                    .where(uninvitedRecruitsCId, '==', fieldMapping.yes)
                    .startAfter(lastDocument)
                    .limit(limit)
                    .get();
            } else {
                break;
            }
        }

        return true;
    } catch (error) {
        console.error(`Error occurred when updating documents: ${error}`);
        throw new Error(error);
    }
};

**Comment by JoeArmani** [comment]
@jhflorey okay, so each key needs to be processed separately? can we just delete the document instead?

**Comment by jhflorey** [comment]
> @jhflorey okay, so each key needs to be processed separately? can we just delete the document instead?

@JoeArmani i did not find it yet. Let me research more

File: utils/firestore.js [hunk 3]
```diff
@@ -453,6 +453,35 @@ const removeParticipantsDataDestruction = async () => {
...
+                batch.delete(doc.ref);
+            }
+        
+            await batch.commit();
+        }
+
+        return true
+    } catch (error) {
+        willContinue = false;
+        console.error(`Error occurred when updating documents: ${error}`);
```
we-ai, line 480:
```suggestion
        console.error(`Error occurred when deleting documents: ${error}`);
```

we-ai (approved):
Since the function will run at scheduled time daily and is not http-triggered, checking http requests and sending responses may not be necessary.

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,24 @@
+const { getResponseJSON } = require('./shared');
+
+const participantDataCleanupAPIs = async (req, res) => {
```
we-ai, line 3:
```suggestion
const participantDataCleanup = async (req, res) => {
```

we-ai: For event-triggered functions, there'll be no http requests passed on.

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,24 @@
...
+    const query = req.query;
+    if (!query.api) return res.status(400).json(getResponseJSON('Bad request!', 400));
+    const api = query.api;
+    console.log(api)
+
+    if (api === 'destruction') {
```
we-ai, line 9:
```suggestion
```

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,24 @@
...
+        const { removeParticipantsDataDestruction } = require(`./firestore`);
+        await removeParticipantsDataDestruction()
+    }
+    else if (api === 'uninvited') {
```
we-ai, line 13:
```suggestion
```

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,24 @@
...
+        const { removeUninvitedParticipants } = require(`./firestore`);
+        await removeUninvitedParticipants()
+    }
+    else return res.status(400).json(getResponseJSON('Bad request!', 400));
+
+    res.status(200).json({ message: 'Success!', code: 200 })
+}
```
we-ai, line 20:
```suggestion
    }
    else return res.status(400).json(getResponseJSON('Bad request!', 400));

    res.status(200).json({ message: 'Success!', code: 200 })
}
```

we-ai (commented):
Looks good. Since the implementation is a scheduled function run, not a http request endpoint, current code needs some updates for the function run.

utils/firestore.js
```diff
@@ -431,25 +432,52 @@ const removeParticipantsDataDestruction = async () => {
                 ? new Date().getTime() - new Date(participant['269050420']).getTime()
                 : 0
 
-            if (participant['883668444'] === 704529432 || (Math.floor(timeDiiff / millisecondsInDay) > 60)) {
+            if (participant['883668444'] === 704529432 || (Math.floor(timeDiiff / millisecondsInDay) > 2)) {
```
Gbarra9, line 435:
```suggestion
            if (participant['883668444'] === 704529432 || (Math.floor(timeDiff / millisecondsInDay) > 2)) {
```

A minor change, I think also the variable name should be changed to [`timeDiff`](https://github.com/episphere/connectFaas/pull/362/files#diff-be91e66430aa4dd2f61668d1bd17d3cb449f57947e13bfa2f6bc087919967282R431).

jhflorey: Solved

utils/firestore.js
```diff
@@ -427,29 +428,56 @@ const removeParticipantsDataDestruction = async () => {
             const participant = doc.data();
             const millisecondsInDay = 24 * 60 * 60 * 1000;
```
we-ai, line 429: This can be moved out of loop. There's no need to calculate it in every iteration.

utils/firestore.js
```diff
@@ -427,29 +428,56 @@ const removeParticipantsDataDestruction = async () => {
...
             const { isIsoDate } = require('./validation');
-            const timeDiiff = isIsoDate(participant['269050420'])
+            const timeDiff = isIsoDate(participant['269050420'])
                 ? new Date().getTime() - new Date(participant['269050420']).getTime()
                 : 0
 
-            if (participant['883668444'] === 704529432 || (Math.floor(timeDiiff / millisecondsInDay) > 60)) {
+            if (participant['883668444'] === 704529432 || (Math.floor(timeDiff / millisecondsInDay) > 2)) {
```
we-ai, line 435:
```suggestion
            if (participant['883668444'] === 704529432 || timeDiff > millisecondsInTwoDays)) {
```
Since milliseconds in 2 days is fixed, there's no need to do the calculation every time.

anthonypetersen:
Please use existing YAML files and don't make new ones yet

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,13 @@
+const participantDataCleanup = async () => {
+
+    console.log(`Start cleaning up participant data`);
+    const { removeParticipantsDataDestruction } = require(`./firestore`);
```
we-ai, line 4: Move the import to top.

utils/participantDataCleanup.js
```diff
@@ -0,0 +1,13 @@
...
+    const { removeUninvitedParticipants } = require(`./firestore`);
```
we-ai, line 5: Move the import to top.

Gbarra9 (changes requested):

Gbarra9 (approved):

utils/participantDataCleanup.js
jhflorey: Solved

utils/participantDataCleanup.js
jhflorey: Solved

utils/participantDataCleanup.js
jhflorey: Solved

utils/firestore.js
jhflorey: Solved

utils/firestore.js
jhflorey: Solved

utils/participantDataCleanup.js
jhflorey: Solved

config/dev/cloudbuild5.yaml
```diff
@@ -15,7 +15,10 @@ steps:
...
   args: ['functions', 'add-iam-policy-binding', 'incentiveCompleted', '--member=allUsers', '--role=${_ROLE}']
+- name: 'gcr.io/cloud-builders/gcloud'
+  args: ['functions', 'deploy', 'participantDataCleanup', '--trigger-topic=${_TOPIC}', '--runtime=${_RUNTIME}', '--timeout=480s', '--source=${_SOURCE}', '--env-vars-file=config/dev/.env.yaml']
 substitutions:
   _SOURCE: https://source.developers.google.com/projects/nih-nci-dceg-connect-dev/repos/github_episphere_connectfaas/moveable-aliases/dev
   _RUNTIME: nodejs14
-  _ROLE: roles/cloudfunctions.invoker
\ No newline at end of file
+  _ROLE: roles/cloudfunctions.invoker
+  _TOPIC: participant-data-cleanup
```
we-ai, line 24: Is this topic set up already in GCP?

jhflorey: Yeah, i created new one.

PR #363, opened 2023-05-25
stage->prod (5-25-2023)

we-ai:
Issues addressed:
- Issue [#493](https://github.com/episphere/dashboard/issues/493) implementations and bug fixes
- Issue [#639](https://github.com/episphere/connect/issues/639)
- Issue [#609](https://github.com/episphere/connect/issues/609)


anthonypetersen:
should we be concerned that so much code got removed from `package-lock.json` ?

anthonypetersen:
also I see that it's underneath an object called `devDependencies` but it looks like code Gene pushed updates a functions-framework version? not sure if that would actually be used within GCP as opposed to locally.

utils/dashboard.js
```diff
@@ -7,34 +8,50 @@ const dashboard = async (req, res) => {
...
     const authObj = await isParentEntity(siteDetails);
     if(userEmail) authObj['userEmail'] = userEmail;
     const isParent = authObj.isParent;
     const siteCodes = authObj.siteCodes;
     const isCoordinatingCenter = authObj.coordinatingCenter;
     const isHelpDesk = authObj.helpDesk;
-    const query = req.query;
-    if(!query.api) return res.status(400).json(getResponseJSON('Bad request!', 400));
-    const api = query.api;
-    console.log(api);
```
anthonypetersen, line 37: in the future.. I kind of like having these console log statements in the API... helps add a little bit of context to the logs when debugging

anthonypetersen (approved):

we-ai:
> also I see that it's underneath an object called `devDependencies` but it looks like code Gene pushed updates a functions-framework version? not sure if that would actually be used within GCP as opposed to locally.

I'm less concerned about 'devDependencies'. These packages should only be used during development.

we-ai:
> should we be concerned that so much code got removed from `package-lock.json` ?

Cloud builds work well.  It should be fine updating packages.

PR #364, opened 2023-06-06
BPTL Dashboard: Enable collection search across all sites for BPTL users

abhinavjonnada82:
This PR addresses following issue https://github.com/episphere/biospecimen/issues/426:
Originally users could lookup only NIH specimens. Moving forward BPTL users can
search for specimens using collection ID from different sites.

Related PR: https://github.com/episphere/biospecimen/pull/536

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -215,7 +215,8 @@ const biospecimenAPIs = async (req, res) => {
             const masterSpecimenId = req.query.masterSpecimenId;
             if(!masterSpecimenId) return res.status(400).json(getResponseJSON('Bad request!', 400));
             const { searchSpecimen } = require('./firestore');
-            const response = await searchSpecimen(masterSpecimenId, siteCode);
+            const allSitesFlag = (req.query.allSitesFlag) ? true : false
+            const response = await searchSpecimen(masterSpecimenId, siteCode, allSitesFlag);
             if(!response) return res.status(404).json(getResponseJSON('Data not found!', 404));
             return res.status(200).json({data: response, code:200});
         }
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -789,8 +789,7 @@ const filterDB = async (queries, siteCode, isParent) => {
             if(key === 'token') query = query.where('token', '==', queries[key]);
             if(key === 'studyId') query = query.where('state.studyId', '==', queries[key]);
         }
-        let snapshot = ``
-        queries['allSiteSearch'] === 'true' ? snapshot = await query.get() : snapshot = await query.where('827220437', operator, siteCode).get();
+        const snapshot = await (queries['allSiteSearch'] === 'true' ? query.get() : query.where('827220437', operator, siteCode).get());
         if(snapshot.size !== 0){
             return snapshot.docs.map(document => document.data());
         }
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -789,7 +789,9 @@ const filterDB = async (queries, siteCode, isParent) => {
             if(key === 'token') query = query.where('token', '==', queries[key]);
             if(key === 'studyId') query = query.where('state.studyId', '==', queries[key]);
         }
-        const snapshot = await query.where('827220437', operator, siteCode).get();
+        let snapshot = ``
+        queries['allSiteSearch'] === 'true' ? snapshot = await query.get() : snapshot = await query.where('827220437', operator, siteCode).get();
+        const snapshot = await query.where('827220437', 'in', siteCodes).get();
         if(snapshot.size !== 0){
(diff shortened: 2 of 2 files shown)


utils/biospecimen.js
```diff
@@ -215,7 +215,8 @@ const biospecimenAPIs = async (req, res) => {
             const masterSpecimenId = req.query.masterSpecimenId;
             if(!masterSpecimenId) return res.status(400).json(getResponseJSON('Bad request!', 400));
             const { searchSpecimen } = require('./firestore');
-            const response = await searchSpecimen(masterSpecimenId, siteCode);
+            const allSitesFlag = (req.query.allSitesFlag) ? true : false
```
JoeArmani, line 218: No changes requested, just a comment:
This shortened version might be worth considering: `const allSitesFlag = !!req.query.allSitesFlag`

utils/firestore.js
```diff
@@ -789,7 +789,9 @@ const filterDB = async (queries, siteCode, isParent) => {
             if(key === 'token') query = query.where('token', '==', queries[key]);
             if(key === 'studyId') query = query.where('state.studyId', '==', queries[key]);
         }
-        const snapshot = await query.where('827220437', operator, siteCode).get();
+        let snapshot = ``
+        queries['allSiteSearch'] === 'true' ? snapshot = await query.get() : snapshot = await query.where('827220437', operator, siteCode).get();
+        const snapshot = await query.where('827220437', operator, siteCodes).get();
```
JoeArmani, line 794: Maybe I am mistaken - Is `snapshot` declared twice here? I see both `let snapshot` and `const snapshot`.
I think we can factor out `snapshot` and `await` in the first part like this (but please test it, I haven't):
`const snapshot = await (queries['allSiteSearch'] === 'true' ? query.get() : query.where('827220437', operator, siteCode).get());`

JoeArmani (commented):
Please let me know if you disagree with anything - my first time looking at this code. Thanks.

utils/firestore.js
abhinavjonnada82: duplicate const snapshot creeped in when i did a git pull & merge. Removed it

utils/firestore.js
```diff
@@ -999,13 +1001,14 @@ const reportMissingSpecimen = async (siteAcronym, requestData) => {
...
+const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
     const snapshot = await db.collection('biospecimen').where('820476880', '==', masterSpecimenId).get();
-    if(snapshot.size === 1) {
+    if (snapshot.size === 1) {
         const token = snapshot.docs[0].data().token;
         const response = await db.collection('participants').where('token', '==', token).get();
         const participantSiteCode = response.docs[0].data()['827220437'];
-        if(participantSiteCode === siteCode) return snapshot.docs[0].data();
+        if (allSitesFlag) return snapshot.docs[0].data();
+        else if (!allSitesFlag && participantSiteCode === siteCode) return snapshot.docs[0].data();
```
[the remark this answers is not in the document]
abhinavjonnada82: reason I added !allSitesFlag is allSitesFlag variable could be set to undefined & thats a very edge case. To prevent that, I had to add  !allSitesFlag safeguard.

utils/biospecimen.js
abhinavjonnada82: I have never used !! let me look into it.

utils/firestore.js
abhinavjonnada82: I like your refactor so Im gonna replace it let with const.

JoeArmani (approved):
Looks good! Just a note: This is my first time looking at this part of connectFaas, so my context is limited.

Gbarra9 (approved):
Looks good. I like your approach in reusing the existing `searchSpecimen` function to search a collection via all sites.

utils/firestore.js
```diff
@@ -999,13 +999,14 @@ const reportMissingSpecimen = async (siteAcronym, requestData) => {
 
 }
 
-const searchSpecimen = async (masterSpecimenId, siteCode) => {
+const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
     const snapshot = await db.collection('biospecimen').where('820476880', '==', masterSpecimenId).get();
-    if(snapshot.size === 1) {
+    if (snapshot.size === 1) {
         const token = snapshot.docs[0].data().token;
```
we-ai, line 1005: If all site search is enabled, further check in `participants` is not necessary, right?
Correct me if I'm mistaken. Below line 1004, the result can be returned, like so:
```js
if (allSitesFlag) return snapshot.docs[0].data();
```

we-ai (approved):
Approval with change suggestion.

PR #365, opened 2023-06-08
Update API Logic - getUserProfile

anthonypetersen:
Addresses the following issues
* https://github.com/episphere/connect/issues/647
* https://github.com/episphere/connect/issues/638

Technical Changes
* removed extra `console.log()` calls from `SSOValidation()`
* added extra info to `console.log()` statements at beginning of `biospecimen` API
* consolidated `getUserProfile` APIs from `biospecimen.js` and `submission.js` into `shared.js`
* modified `retrieveUserProfile` to always return an object unless an error occurs
* added shared function `isEmpty` to check whether an object is empty or not

NOTE: Coordinated changes are required to Biospecimen Dashboard and possibly PWA

### DIFF
--- a/utils/shared.js
+++ b/utils/shared.js
@@ -706,8 +706,6 @@ const getUserProfile = async (req, res, uid) => {
         return res.status(500).json(getResponseJSON(responseProfile.message, 500));
     }
 
-    console.log(responseProfile);
-
     if(!isEmpty(responseProfile)){
 
         let responseDefaults = await checkDefaultFlags(responseProfile, uid);


JoeArmani (approved):
√-Add logs
√-check uid exists in query
√-return participant at index 0 instead of .map()
√-retrieveUserProfile either returns a user or throws an error
√-errors caught in getUserProfile()
√-iterate and update default values if needed
√-return responseProfile && status 200 or first realized error code

looks good

utils/shared.js
```diff
@@ -693,6 +693,51 @@ const swapObjKeysAndValues = (object) => {
...
+    }
+
+    const { retrieveUserProfile } = require('./firestore');
+    let responseProfile = await retrieveUserProfile(uid);
+
+    if(responseProfile instanceof Error){
+        return res.status(500).json(getResponseJSON(responseProfile.message, 500));
+    }
+
+    console.log(responseProfile);
```
we-ai, line 709: A bit context about the log will be helpful, like other updated log messages.

we-ai (approved):
Looks good. Commented on minor things.

we-ai (approved):

utils/shared.js
anthonypetersen: This shouldn't be there, it was from troubleshooting... Removing

jhflorey (approved):

PR #368, opened 2023-06-09
Return code 200 when a collection is not found

we-ai:
This PR address issue [#628](https://github.com/episphere/connect/issues/628). Below are details:
- Updated `searchSpecimen` API. It now returns response code 200 and an empty data object, when a biospecimen search returns no record
- Fixed a typo and add back a log in `dashboard.js`

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -215,9 +215,14 @@ const biospecimenAPIs = async (req, res) => {
             const { searchSpecimen } = require('./firestore');
             const masterSpecimenId = req.query.masterSpecimenId;
             const allSitesFlag = (req.query.allSitesFlag) ? true : false;
-            const response = await searchSpecimen(masterSpecimenId, siteCode, allSitesFlag);
-            if (!response) return res.status(200).json({data: {}, code:200});
-            return res.status(200).json({data: response, code:200});
+
+            try {
+              const biospecimenData = await searchSpecimen(masterSpecimenId, siteCode, allSitesFlag);
+              return res.status(200).json({ data: biospecimenData, code: 200 });
+            } catch (error) {
+              console.error('Error occurred when running searchSpecimen:', error);
+              return res.status(500).json({ message: 'Error occurred when running searchSpecimen.', code: 500 });
+            }
         }
         else {
             const { searchShipments } = require('./firestore');
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -218,7 +218,7 @@ const biospecimenAPIs = async (req, res) => {
 
             try {
               const biospecimenData = await searchSpecimen(masterSpecimenId, siteCode, allSitesFlag);
-              return res.status(200).json({ data: biospecimenData, code: 200 });
+              return res.status(200).json({ data: biospecimenData, message: 'Success!', code: 200 });
             } catch (error) {
               console.error('Error occurred when running searchSpecimen:', error);
               return res.status(500).json({ data: {}, message: 'Error occurred when running searchSpecimen.', code: 500 });
@@ -242,7 +242,7 @@ const biospecimenAPIs = async (req, res) => {
 
             try {
               const specimenArray = await getSpecimenCollections(token, siteCode);
-              return res.status(200).json({ data: specimenArray, code: 200 });
+              return res.status(200).json({ data: specimenArray, message: 'Success!', code: 200 });
             } catch (error) {
               console.error('Error occurred when running getSpecimenCollections:', error);
               return res.status(500).json({ data: [], message: 'Error occurred when running getSpecimenCollections.', code: 500 });
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -237,14 +237,18 @@ const biospecimenAPIs = async (req, res) => {
             return res.status(405).json(getResponseJSON('Only GET requests are accepted!', 405));
         }
         if(req.query.token) {
-            const token = req.query.token;
-            if(!token) return res.status(400).json(getResponseJSON('Bad request!', 400));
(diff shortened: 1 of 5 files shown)


utils/biospecimen.js
```diff
@@ -212,12 +212,11 @@ const biospecimenAPIs = async (req, res) => {
...
         if(req.query.masterSpecimenId) {
-            const masterSpecimenId = req.query.masterSpecimenId;
-            if(!masterSpecimenId) return res.status(400).json(getResponseJSON('Bad request!', 400));
             const { searchSpecimen } = require('./firestore');
-            const allSitesFlag = (req.query.allSitesFlag) ? true : false
+            const masterSpecimenId = req.query.masterSpecimenId;
+            const allSitesFlag = (req.query.allSitesFlag) ? true : false;
             const response = await searchSpecimen(masterSpecimenId, siteCode, allSitesFlag);
-            if(!response) return res.status(404).json(getResponseJSON('Data not found!', 404));
+            if (!response) return res.status(200).json({data: {}, code:200});
```
anthonypetersen, line 219: might suggest combining the `200` return statements into one line that is always going to return something for data and code, then let `searchSpecimen` handle either returning a populated object or an empty object. Also would suggest having error handling in `searchSpecimen` and then checking here `if(response instanceof Error)`

anthonypetersen (changes requested):
while this does handle another instance of our API returning a `404` code for a call that had no errors, the Issue this PR calls out was meant to handle two different APIs that eventually call `getSpecimenCollections()` in `firestore.js`

* Biospecimen Dashboard API `getParticipantCollections`
* PWA API `getUserCollections`

utils/biospecimen.js
```diff
@@ -212,12 +212,11 @@ const biospecimenAPIs = async (req, res) => {
             return res.status(405).json(getResponseJSON('Only GET requests are accepted!', 405));
         }
         if(req.query.masterSpecimenId) {
-            const masterSpecimenId = req.query.masterSpecimenId;
-            if(!masterSpecimenId) return res.status(400).json(getResponseJSON('Bad request!', 400));
             const { searchSpecimen } = require('./firestore');
```
jhflorey, line 215: Should we put `const { searchSpecimen } = require('./firestore');` at the top of the file?

utils/biospecimen.js
```diff
@@ -212,12 +212,11 @@ const biospecimenAPIs = async (req, res) => {
...
-            const allSitesFlag = (req.query.allSitesFlag) ? true : false
+            const masterSpecimenId = req.query.masterSpecimenId;
```
jhflorey, line 216: NIT: We could write `const { masterSpecimenID } = req.query` instead

utils/biospecimen.js
jhflorey: I have a question if `searchSpecimen` function has an internal error what will be the result of `response`?

utils/biospecimen.js
abhinavjonnada82: I think it makes more sense to leave it on line 215. You can initialize` const { searchSpecimen } = require('./firestore'); `only `if req.query.masterSpecimenId` is `true`. This saves some computation instead of initializing it at the very beginning.

utils/biospecimen.js
abhinavjonnada82: `const masterSpecimenId = req.query.masterSpecimenId;`
Above approach is a direct assignment. Your suggested way is useful only if we have multiple properties from an object.
@we-ai do you think any one of them has better performance?

utils/biospecimen.js
we-ai: Combined returns, and added error handler.

utils/biospecimen.js
we-ai: Direct assignment is slightly faster, but the performance differences are minimal. I tested on a small object: `const obj ={ a: { ab: { abc: 1 } }, b: 2, c: 3, d: 4, e: 5 }`. Tested on 1 billion operations on `const {abc} = obj.a.ab` and `const abc = obj.a.ab.abc` . Below are 10 test runs if you're interested:
```
[
  [ 'Data Destruction: 588ms', 'Direct Assignment: 564ms' ],
  [ 'Data Destruction: 609ms', 'Direct Assignment: 584ms' ],
  [ 'Data Destruction: 570ms', 'Direct Assignment: 521ms' ],
  [ 'Data Destruction: 555ms', 'Direct Assignment: 544ms' ],
  [ 'Data Destruction: 516ms', 'Direct Assignment: 498ms' ],
  [ 'Data Destruction: 509ms', 'Direct Assignment: 494ms' ],
  [ 'Data Destruction: 505ms', 'Direct Assignment: 489ms' ],
  [ 'Data Destruction: 496ms', 'Direct Assignment: 494ms' ],
  [ 'Data Destruction: 481ms', 'Direct Assignment: 500ms' ],
  [ 'Data Destruction: 479ms', 'Direct Assignment: 481ms' ]
]
```

we-ai:
> 
The `searchSpecimen` function in "firestore.js" is only used in `api === 'searchSpecimen'` in "biospecimen.js". PWA is not using this API and function.

The `getUserCollections` API in PWA calls `getSpecimenCollections` function, which is a separate function in "firestore.js".

utils/biospecimen.js
abhinavjonnada82: Thanks @we-ai

abhinavjonnada82 (approved):
Looks good nothing to suggest & thanks @we-ai for addressing those comments!

utils/biospecimen.js
we-ai: We're using conditional imports in API calls, and we can discuss its pros and cons in our use cases.

PR #369, opened 2023-06-12
Bug Fix - participant-data-cleanup job removes stub records.

jhflorey:
This PR continue to address a issue that SMDB can not look up participant who removed data destruction

- Renamed `stubRecordList` to `stubFieldArray` as request from @we-ai 
- Added `699625233 : Autogenerated flag when User Profile submitted` to `stubFieldArray`
- Recounted the number of participants whose data has been delete

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -417,7 +417,7 @@ const removeParticipantsDataDestruction = async () => {
     try {
         let count = 0;
         const millisecondsInTwoDays = 2 * 24 * 60 * 60 * 1000;
-        const stubRecordList = ['104278817', '119449326', '153713899', '173836415', '231676651', '262613359', '268665918', '269050420', '304438543', '359404406', '399159511', '407743866', '412000022', '471168198', '479278368', '524352591', '526455436', '544150384', '558435199', '577794331', '592227431', '613641698', '664453818', '744604255', '747006172', '765336427', '773707518', '826240317', '831041022', '883668444', '996038075', 'token', 'Connect_ID', 'query', 'pin', 'state', '371067537', '827220437'];
+        const stubFieldArray = ['104278817', '119449326', '153713899', '173836415', '231676651', '262613359', '268665918', '269050420', '304438543', '359404406', '399159511', '407743866', '412000022', '471168198', '479278368', '524352591', '526455436', '544150384', '558435199', '577794331', '592227431', '613641698', '664453818', '744604255', '747006172', '765336427', '773707518', '826240317', '831041022', '883668444', '996038075', 'token', 'Connect_ID', 'query', 'pin', 'state', '371067537', '827220437', '699625233'];
         const destroyDataCId = fieldMapping.participantMap.destroyData.toString();
         const dateRequestedDataDestroyCId = fieldMapping.participantMap.dateRequestedDataDestroy.toString();
         const destroyDataCategoricalCId = fieldMapping.participantMap.destroyDataCategorical.toString();
@@ -437,14 +437,16 @@ const removeParticipantsDataDestruction = async () => {
                 : 0
 
             if (participant[destroyDataCategoricalCId] === requestedAndSignCId || timeDiff > millisecondsInTwoDays) {
+                let hasRemovedField = false;
                 const fieldKeys = Object.keys(participant)
                 const participantRef = doc.ref;
                 fieldKeys.forEach(key => {
-                    if (!stubRecordList.includes(key)) {
+                    if (!stubFieldArray.includes(key)) {
                         batch.update(participantRef, { [key]: admin.firestore.FieldValue.delete() });
+                        hasRemovedField = true
                     }
                 })
-                count++;
+                if (hasRemovedField) count++;
             }
             await batch.commit()
         }


utils/firestore.js
```diff
@@ -437,14 +437,16 @@ const removeParticipantsDataDestruction = async () => {
...
                 const participantRef = doc.ref;
                 fieldKeys.forEach(key => {
-                    if (!stubRecordList.includes(key)) {
+                    if (!stubFieldArray.includes(key)) {
                         batch.update(participantRef, { [key]: admin.firestore.FieldValue.delete() });
+                        hasRemovedField = true
                     }
                 })
-                count++;
+                if (hasRemovedField) count++;
```
abhinavjonnada82, line 449: Out of curiosity are you using count anywhere else?

JoeArmani (approved):
Approved with some comments - I'm not totally clear about the bounds/scale of when this function is called and if it will ever approach 500 operations it should be in a loop with pagination. 

-maybe `hasRemovedField` could be omitted and the `counter` could be incremented instead of `hasFieldRemoved = true`?

-can the `isIsoDate require` statement be moved out of the for loop?

-does this ever need to handle more than 500 docs as the study grows? If yes, we'll need the routine inside a loop.

-NIT: it probably makes sense to hardcode millisecondsInTwoDays instead of calculating it every time since that value won't change

abhinavjonnada82:
> -NIT: it probably makes sense to hardcode millisecondsInTwoDays instead of calculating it every time since that value won't change

I had the same doubt, I think the value is large but still can save computation cost if you hardcode the value as Joe mentioned
`const millisecondsInTwoDays = 172800000 //  2 * 24 * 60 * 60 * 1000`

utils/firestore.js
```diff
@@ -437,14 +437,16 @@ const removeParticipantsDataDestruction = async () => {
...
 
             if (participant[destroyDataCategoricalCId] === requestedAndSignCId || timeDiff > millisecondsInTwoDays) {
+                let hasRemovedField = false;
                 const fieldKeys = Object.keys(participant)
                 const participantRef = doc.ref;
                 fieldKeys.forEach(key => {
-                    if (!stubRecordList.includes(key)) {
+                    if (!stubFieldArray.includes(key)) {
                         batch.update(participantRef, { [key]: admin.firestore.FieldValue.delete() });
+                        hasRemovedField = true
```
abhinavjonnada82, line 446: Maybe swap `count++` with `hasRemovedField = true` and get rid of `hasRemovedField`

jhflorey:
@JoeArmani @abhinavjonnada82 Already updated 

- can the isIsoDate require statement be moved out of the for loop?
- NIT: it probably makes sense to hardcode millisecondsInTwoDays instead of calculating it every time since that value won't change

utils/firestore.js
```diff
@@ -431,20 +433,21 @@ const removeParticipantsDataDestruction = async () => {
...
                     }
                 })
-                count++;
+                if (hasRemovedField) count++;
             }
             await batch.commit()
```
we-ai, line 452: An error message will be helpful when the batch commit fails.

JoeArmani:
> @JoeArmani @abhinavjonnada82
> 
> `maybe hasRemovedField could be omitted and the counter could be incremented instead of hasFieldRemoved = true?` => The counter is for number of participants and not for the number of fields. `does this ever need to handle more than 500 docs as the study grows? If yes, we'll need the routine inside a loop.` => I do not think it will be reach to 500docs since the job run every day to remove participant data destruction.

Okay so I have a little more understanding now. Thanks @jhflorey. This runs for the keys the participant has requested destruction on and batch commit() is executed for each participant. The counter is used for logging only (participant count, not key count). And this process makes sense when a participant has requested data destruction but not removal (which would be a document delete).

I tend to be in favor of wrapping the whole thing in a while loop with a get() limit =500 and pagination just for scalability, but I agree that it's very unlikely this will ever process 500+ participants in a day based on the scale of the study.

Looks good.

abhinavjonnada82 (approved):

utils/firestore.js
jhflorey: We already had it at `catch` statement
image

jhflorey:
> > @JoeArmani @abhinavjonnada82
> > `maybe hasRemovedField could be omitted and the counter could be incremented instead of hasFieldRemoved = true?` => The counter is for number of participants and not for the number of fields. `does this ever need to handle more than 500 docs as the study grows? If yes, we'll need the routine inside a loop.` => I do not think it will be reach to 500docs since the job run every day to remove participant data destruction.
> 
> Okay so I have a little more understanding now. Thanks @jhflorey. This runs for the keys the participant has requested destruction on and batch commit() is executed for each participant. The counter is used for logging only (participant count, not key count). And this process makes sense when a participant has requested data destruction but not removal (which would be a document delete).
> 
> I tend to be in favor of wrapping the whole thing in a while loop with a get() limit =500 and pagination just for scalability, but I agree that it's very unlikely this will ever process 500+ participants in a day based on the scale of the study.
> 
> Looks good.

@JoeArmani  you are correct. We only remove some fields (except stub records), not remove whole data (document). BTW thanks for your feedbacks and suggestions

we-ai (approved):

PR #370, opened 2023-06-15
SiteManager Dashboard: Add updated User Profile F + L name into search query 

abhinavjonnada82:
Title^^
User can look up participants on updated user profile first or last name. This PR addresses the following issue:
https://github.com/episphere/dashboard/issues/540

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -108,10 +108,6 @@ const updateResponse = async (data, uid) => {
         const response = await db.collection('participants').where('state.uid', '==', uid).get();
         if(response.size === 1) {
             for(let doc of response.docs){
-                if (data[399159511] || data[996038075]) { 
-                    const query = updateUPSearchQuery(data, doc.data().query);
-                    data = { ...data, query };
-                }
                 await db.collection('participants').doc(doc.id).update(data);
                 return true;
             }
@@ -121,16 +117,6 @@ const updateResponse = async (data, uid) => {
         console.error(error);
         return new Error(error)
     }
-}
-
-const updateUPSearchQuery = (data, previousQuery) => {
-    const firstName = data[399159511]?.trim()?.toLowerCase();
-    const lastName = data[996038075]?.trim()?.toLowerCase();
-    return {
-        ...previousQuery,
-        ...(firstName && { firstName }),
-        ...(lastName && { lastName })
-    };
 }
 
 const incrementCounter = async (field, siteCode) => {
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -103,6 +103,7 @@ const participantExists = async (uid) => {
     }
 }
 
+
 const updateResponse = async (data, uid) => {
     try{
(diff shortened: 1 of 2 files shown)


utils/firestore.js
```diff
@@ -109,6 +109,25 @@ const updateResponse = async (data, uid) => {
         const response = await db.collection('participants').where('state.uid', '==', uid).get();
         if(response.size === 1) {
             for(let doc of response.docs){
+                if(data[399159511] && data[996038075]) { // if user profile F+L name is updated
+                    const previousQuery = doc.data().query
+                    const firstName = data[399159511].trim().toLowerCase()
```
anthonypetersen, line 114: I would suggest moving this logic to `submit()` instead of inside this function... Also please condense this code, there are multiple statements made over and over. Could be much simpler.

anthonypetersen (changes requested):

utils/firestore.js
abhinavjonnada82: My bad just optimized. I need access to users`doc.data()` to access previous query that's the reason I wrote the code here instead of `submit()`.

utils/firestore.js
```diff
@@ -109,6 +109,16 @@ const updateResponse = async (data, uid) => {
         const response = await db.collection('participants').where('state.uid', '==', uid).get();
         if(response.size === 1) {
             for(let doc of response.docs){
+                const previousQuery = doc.data().query;
+                const firstName = data[399159511]?.trim()?.toLowerCase();
+                const lastName = data[996038075]?.trim()?.toLowerCase();
```
anthonypetersen, line 114: see other comment and please move outside of this function

anthonypetersen (changes requested):

utils/firestore.js
abhinavjonnada82: moved to submission.js

utils/submission.js
```diff
@@ -73,6 +73,19 @@ const submit = async (res, data, uid) => {
         }
     }
 
+    if (data[399159511] || data[996038075]) { // update user profile F + L name in search query
+        const { retrieveUserProfile } = require('./firestore');
+        const previousQuery = (await retrieveUserProfile(uid))[0].query;
+        const firstName = data[399159511]?.trim()?.toLowerCase();
```
anthonypetersen, line 79: This seems like a complicated way of doing things...

if(data[399159511]) data['query.firstName'] = data[399159511].trim().toLowerCase();
if(data[996038075]) data['query.lastName'] = data[996038075].trim().toLowerCase();

anthonypetersen (changes requested):
see comments

utils/submission.js
abhinavjonnada82: your suggested way of code replaces the key-value pairs in query object instead of appending to the object. 
I think its concise

abhinavjonnada82:
This changes would be handled in this 
PR: https://github.com/episphere/dashboard/pull/538/files

PR #381, opened 2023-07-10
Expand name search for query.firstName and query.lastName arrays

JoeArmani:
This PR is in support of https://github.com/episphere/connect/issues/654 - expand name search.

Update the `filterDb()` function to handle both string and array<string> fields for participant search fields `query.firstName` and `query.lastName`.

If `name` is included in the `query`, we have to handle two cases: sometimes `typeof(participantDoc.query.firstName) === 'string'` and sometimes `typeof(participantDoc.query.firstName) === 'array'`.
•If it's a string, we can just do a simple `query.where('query.firstName', '==', queries.firstName.toLowerCase())`.
•If it's an array, we have to do a `query.where('query.firstName', 'array-contains', queries.firstName.toLowerCase())`.

The caveat to `array-contains` searches: we can only have one `array-contains` per query. So, when multiple `array-contains` items are present (such as firstName, lastName, email, and/or phone), we need to process those queries separately and filter the results.

When any version of `query.firstName` or `query.lastName` is present, we need to check for both the string version and the array version, then filter the results.

When neither `query.firstName` nor `query.lastName` is present, we can run a simple query as we did before. This returns early and skips filtering operations (additional filtering is not necessary in this case).

Examples:
http://127.0.0.1:5001/nih-nci-dceg-connect-dev/us-central1/dashboard?api=getParticipants&type=filter&lastName=tree&dob=19600101

>  queries.lastName tree
>  queries.dob 19600101
>  filteredResults size: 2
>  filtered participant 1: ophelia tree, Connect_ID: 6379614904
>  filtered participant 2: owen tree, Connect_ID: 3618260663

http://127.0.0.1:5001/nih-nci-dceg-connect-dev/us-central1/dashboard?api=getParticipants&type=filter&firstName=homer

>  queries.firstName homer
>  filteredResults size: 3
>  filtered participant 1: homer simpson, Connect_ID: 7139599412
>  filtered participant 2: homer simpson, Connect_ID: 5068657761
>  filtered participant 3: homer simpson, Connect_ID: 6471218248

http://127.0.0.1:5001/nih-nci-dceg-connect-dev/us-central1/dashboard?api=getParticipants&type=filter&connectId=3503670213

>  { api: 'getParticipants', connectId: '3503670213' }
>  snapshot size - [object Object]: 1
>  filtered participant 1: kevins,k,kevin levin,lavin, Connect_ID: 3503670213

http://127.0.0.1:5001/nih-nci-dceg-connect-dev/us-central1/dashboard?api=getParticipants&type=filter&firstName=lisa&lastName=simpson

>  { api: 'getParticipants', firstName: 'lisa', lastName: 'simpson' }
>  filteredResults size: 2
>  filtered participant 1: lisa simpson, Connect_ID: 7967651258
>  filtered participant 2: lisa simpson, Connect_ID: 8933919034

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -790,161 +790,151 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
  * TODO Future: if we (1) adjust signup to write query.firstName and query.lastName as arrays, and (2) update existing participant docs, we can simplify this function by removing the 'string' searches for firstName and lastName.
  */
 const filterDB = async (queries, siteCode, isParent) => {
-    try {
+
+    // We need to make separate get requests for each query, since Firestore only allows one 'array-contains' query per request.
+    // This function isolates a single array-contains query (firstName, lastName, email, phone).
+    const updateQueryForArrayContainsConstraints = (queryInput) => {
+        const newQueriesObj = {...queries};
+
+        for(let property in newQueriesObj) {
+            if(queryInput.includes(newQueriesObj[property])) {
+                delete newQueriesObj[property];
+            }
+        }
+
+        return newQueriesObj;
+    };
+
+    // This function directs the generation and execution of queries based on the search properties that are present.
+    // If neither firstName nor lastName are present, this function is bypassed.
+    // TODO Future: once all query.firstName and query.lastName are arrays, we can simplify this function by removing the 'string' searches for firstName and lastName.
+    const handleNameQueries = async (firstNameQuery, lastNameQuery, phoneEmailQuery) => {
+        if (firstNameQuery) {
+            const fNameStringQueryForSearch = generateQuery(firstNameQuery, 'string');
+            await executeQuery(fNameStringQueryForSearch);
+
+            const fNameArrayQueryForSearch = generateQuery(firstNameQuery, 'array');
+            await executeQuery(fNameArrayQueryForSearch);
+        }
+
+        if (lastNameQuery) {
+            const lNameStringQueryForSearch = generateQuery(lastNameQuery, 'string');
+            await executeQuery(lNameStringQueryForSearch);
+
+            const lNameArrayQueryForSearch = generateQuery(lastNameQuery, 'array');
+            await executeQuery(lNameArrayQueryForSearch);
+        }
+
+        if (phoneEmailQuery) {
+            const phoneOrEmailQueryForSearch = generateQuery(phoneEmailQuery);
+            await executeQuery(phoneOrEmailQueryForSearch);
+        }
+    };
+
+    // This function generates a query based on the property that are passed in
+    // nameType is either 'string' or 'array' and it only applies to firstName and lastName properties in search.
+    // nameType is currently a string when a participant signs up, but is written as array when a name update is made in connectApp or SMDB as of the PRs on (JA 07/07/23).
+    const generateQuery = (queryKeys, nameType) => {
+        let participantQuery = collection;
+
+        for (let key in queryKeys) {
+            if (key === 'firstName' || key === 'lastName') {
+                const path = `query.${key}`;
+                const queryValue =(key === 'firstName' ? queries.firstName : queries.lastName).toLowerCase();
+                const operation = (nameType === 'string') ? '==' : 'array-contains';
+                participantQuery = participantQuery.where(path, operation, queryValue);
+            }
+            if (key === 'email' || key === 'phone') {
+                const path = `query.${key === 'email' ? 'allEmails' : 'allPhoneNo'}`;
+                const queryValue = (key === 'email' ? queries.email : queries.phone).toLowerCase();
+                participantQuery = participantQuery.where(path, 'array-contains', queryValue);
+            }
+            if (key === 'dob') participantQuery = participantQuery.where('371067537', '==', queries.dob);
+            if (key === 'connectId') participantQuery = participantQuery.where('Connect_ID', '==', parseInt(queries.connectId));
+            if (key === 'token') participantQuery = participantQuery.where('token', '==', queries.token);
+            if (key === 'studyId') participantQuery = participantQuery.where('state.studyId', '==', queries.studyId);
+            if (key === 'checkedIn') participantQuery = participantQuery.where('331584571.266600170.135591601', '==', 353358909);
+        }
+
+        return participantQuery;
+    }   
+
+    // This function executes each query and pushes the data to the fetchedResults array
+    const executeQuery = async (query) => {
         const operator = isParent ? 'in' : '==';
-        const collection = db.collection('participants');
-        let fetchedResults = [];
-
-        // We need to make separate get requests for each query, since Firestore only allows one 'array-contains' query per request.
-        // This function isolates a single array-contains query (firstName, lastName, email, phone).
-        const updateQueryForArrayContainsConstraints = (queryInput) => {
-            const newQueriesObj = {...queries};
-
-            for(let property in newQueriesObj) {
-                if(queryInput.includes(newQueriesObj[property])) {
-                    delete newQueriesObj[property];
-                }
-            }
-
-            return newQueriesObj;
-        };
-
-        // This function directs the generation and execution of queries based on the search properties that are present.
-        // If neither firstName nor lastName are present, this function is bypassed.
-        // TODO Future: once all query.firstName and query.lastName are arrays, we can simplify this function by removing the 'string' searches for firstName and lastName.
-        const handleQueries = async (firstNameQuery, lastNameQuery, phoneEmailQuery) => {
-            if (firstNameQuery) {
-                const fNameStringQueryForSearch = generateQuery(firstNameQuery, 'string');
-                await executeQuery(fNameStringQueryForSearch);
-
-                const fNameArrayQueryForSearch = generateQuery(firstNameQuery, 'array');
-                await executeQuery(fNameArrayQueryForSearch);
-            }
-
-            if (lastNameQuery) {
-                const lNameStringQueryForSearch = generateQuery(lastNameQuery, 'string');
-                await executeQuery(lNameStringQueryForSearch);
-
-                const lNameArrayQueryForSearch = generateQuery(lastNameQuery, 'array');
-                await executeQuery(lNameArrayQueryForSearch);
-            }
-
-            if (phoneEmailQuery) {
-                const phoneOrEmailQueryForSearch = generateQuery(phoneEmailQuery);
-                await executeQuery(phoneOrEmailQueryForSearch);
-            }
-        };
-
-        // This function generates a query based on the property that are passed in
-        // nameType is either 'string' or 'array' and it only applies to firstName and lastName properties in search.
-        // nameType is currently a string when a participant signs up, but is written as array when a name update is made in connectApp or SMDB as of the PRs on (JA 07/07/23).
-        const generateQuery = (queryInput, nameType) => {
-            for (let key in queries) {
-                if (key === 'firstName' || key === 'lastName') {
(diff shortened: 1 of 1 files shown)


utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
         const operator = isParent ? 'in' : '==';
-        let query = db.collection('participants');
-        for(let key in queries) {
-            if(key === 'firstName' || key === 'lastName') query = query.where(`query.${key}`, '==', queries[key].toLowerCase());
-            if(key === 'email' || key === 'phone') query = query.where(`${key === 'email' ? `query.allEmails` : `query.allPhoneNo`}`, 'array-contains', queries[key].toLowerCase());
-            if(key === 'dob') query = query.where('371067537', '==', queries[key]);
-            if(key === 'connectId') query = query.where('Connect_ID', '==', parseInt(queries[key]));
-            if(key === 'token') query = query.where('token', '==', queries[key]);
-            if(key === 'studyId') query = query.where('state.studyId', '==', queries[key]);
-            if(key === 'checkedIn') query = query.where('331584571.266600170.135591601', '==', 353358909);
```
abhinavjonnada82, line 794: Missing `checkedIn` key from the list of queries

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+        } else if (queries.firstName) {
+            await handleQueries(queries, null, null);
+        } else if (queries.lastName) {
+            await handleQueries(null, queries, null);
+        } else {
+            const nonNameQuery = generateQuery(collection, queries);
+            await executeQuery(nonNameQuery);
+
+            return fetchedResults;
         }
```
abhinavjonnada82, line 951: Move this to a separate function

abhinavjonnada82:
@JoeArmani 
I have reviewed the code changes you made to filterDB. I highly suggest that you do not change existing functions or introduce new code into filterDB. filterDB is used across multiple applications and has been working fine for several years. Any new code introduced is a liability for the occurrence of bugs or issues in the future.

Instead, I recommend that you leave filterDB as it is and define your functions outside of filterDB. You can then call these functions from filterDB. This will allow you to add new functionality without introducing any risk to the existing code.

Also, if the code is no longer meeting the needs of the application or there is a security vulnerability. In these cases, feel free to change the function as per requirements.

@we-ai or @anthonypetersen feel free to add to or oppose my comment.

abhinavjonnada82 (changes requested):

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+                return typeof value === 'string'
+                    ? value === queryValue
+                    : Array.isArray(value) && value.includes(queryValue);
+            },
+            email: (participant, queryValue) => participant.query.allEmails.includes(queryValue),
+            phone: (participant, queryValue) => participant.query.allPhoneNo.includes(queryValue),
+            dob: (participant, queryValue) => participant['371067537'] === queryValue,
+        };
+
+        // This function removes results that don't match the query. For example, if the query is for 'John Smith' and the results include 'John Doe', 'John Smith', and 'Jane Smith', we need to remove 'John Doe' and 'Jane Smith' from the results.
```
abhinavjonnada82, line 903: I agree these descriptive notes are helpful but I suggest you condense them 1-2 lines or just remove it entirely.

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
     }
 }
 
+/**
+ * This function is used to retrieve a list of participants from the database.
```
abhinavjonnada82, line 784: Descriptive but condense them to 1-2 lines or just remove it entirely.

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+                    fetchedResults.push(doc.data());
+                });
+            }
+        };
+
+        // This function removes duplicate Connect_ID entries from fetchedResults. Since we have to perform multiple get() requests to handle the array-contains queries,
+        // We expect to end up with duplicate entries in fetchedResults that need to be handled.
+        // Example: A search for 'John Smith' will result in two get() requests: one for 'John' and one for 'Smith'. Both requests will return the same participant, so we need to remove the duplicate.
+        const removeDuplicateResults = () => {
+            let uniques = new Set();
```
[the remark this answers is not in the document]
JoeArmani: Done

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+
+            fetchedResults = fetchedResults.filter(participant => {
+                if (uniques.has(participant.Connect_ID)) {
```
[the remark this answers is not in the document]
JoeArmani: I don’t see the need for a variable here. This returns:
`false` if the `uniqueParticipants Set()` already has the matching Connect_ID. False refers to the filter() function, so the participant is not added to `fetchedResults []`.
`true` if the `uniqueParticipants Set()` does not already have the matching Connect_ID. True refers to the filter() function which includes that participant in `fetchedResults []`.

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+                    return false;
+                } else {
+                    uniques.add(participant.Connect_ID);
+                    return true;
+                }
+            });
+        };
+
+        // Map from query type, returns true if a participant doc should be kept in fetchedResults
+        const queryToResultsChecker = {
```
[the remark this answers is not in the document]
JoeArmani: Updated. The new version uses a compound filter, is more readable, and is performant.

```javascript
const removeMismatchedResults = () => {
        fetchedResults = fetchedResults.filter(participant => {
            return (!queries.firstName || participant.query.firstName.includes(queries.firstName)) &&
                (!queries.lastName || participant.query.lastName.includes(queries.lastName)) &&
                (!queries.email || participant.query.allEmails.includes(queries.email)) &&
                (!queries.phone || participant.query.allPhoneNo.includes(queries.phone)) &&
                (!queries.dob || participant['371067537'] === queries.dob) &&
                (!queries.connectId || participant.Connect_ID === parseInt(queries.connectId)) &&
                (!queries.token || participant.token === queries.token) &&
                (!queries.studyId || participant.state.studyId === queries.studyId) &&
                (!queries.checkedIn || participant['331584571.266600170.135591601'] === 353358909);
        });
    };
```

utils/firestore.js
```diff
@@ -780,32 +780,186 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+            dob: (participant, queryValue) => participant['371067537'] === queryValue,
+        };
+
+        // This function removes results that don't match the query. For example, if the query is for 'John Smith' and the results include 'John Doe', 'John Smith', and 'Jane Smith', we need to remove 'John Doe' and 'Jane Smith' from the results.
+        // This is necessary because we have to run multiple queries to handle the array-contains constraints, so we inevitably end up with results that don't match the query when more than one array-contains operation is executed.
+        // Array-contains reminder: a query can only contain one array-contains operation, so we have to run multiple queries to handle multiple array-contains constraints (see participant query properties: firstName, lastName, email, phone).
+        // For each query type, use the corresponding checker function 
+        const removeMismatchedResults = () => {          
+            for (const queryType in queries) {
+                if (queries.hasOwnProperty(queryType)) {
```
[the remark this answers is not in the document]
JoeArmani: Updated to `Object.keys(queries).includes(queryType)`

utils/firestore.js
JoeArmani: I prefer this flow because it allows us to return early without additional execution or function calls. Let's leave this open for discussion in leads approval.

utils/firestore.js
JoeArmani: I took a pass and shortened all the notes. More will be removed when we complete the changeover to only array structure for `query.firstName` and `query.lastName`.

JoeArmani:
> @JoeArmani I have reviewed the code changes you made to filterDB. I highly suggest that you do not change existing functions or introduce new code into filterDB or other code. filterDB is used across multiple applications and has been working fine for several years. Any new code introduced is a liability for the occurrence of bugs or issues in the future.
> 
> Instead, I recommend that you leave filterDB as it is and define your functions outside of filterDB. You can then call these functions from filterDB. This will allow you to add new functionality without introducing any risk to the existing code.
> 
> Also, if the code is no longer meeting the needs of the application or there is a security vulnerability. In these cases, feel free to change the function as per requirements.
> 
> @we-ai or @anthonypetersen feel free to add to or oppose my comment.

I hear you. The old version of `filterDB()` is still wrapped in this new version. The functionality has just been extended due to the array-contains constraints on `query.firstName` and `query.lastName`. The old functionality is identical for searches that don't include names.

My take is that pointing to other functions, and then back to this one depending on the query would get messy. Changing the caller flow through the apps and connectFaas would be even messier. The other issue is that the previous `filterDB` function is useless with the new array structures (it would fail for any search including names). Because of these items, I believe extending this function was the best way to go for conciseness, maintenance, and extensibility. Let's leave this open to discussion during leads approval because I understand that it may be worth discussing.

abhinavjonnada82 (approved):

anthonypetersen (approved):
looks good

anthonypetersen:
With the structure of `query` changing as dramatically as it is, I don't think it's unexpected that we would want to update `filterDB()`. It feels like a six in one, half a dozen in the other type situation. We obviously want to keep the endpoint "looking the same" from the point of view of the parts of our codebase that call `filterDB()`, but what difference does it make if that function calls other functions within itself or outside the function? The function is still fundamentally getting overhauled, so it feels like a matter of style preference. Unless we think these new functions will also be called elsewhere, I really see no issue in defining them within `filterDB()`

utils/firestore.js
```diff
@@ -780,32 +780,168 @@ const retrieveSiteNotifications = async (siteId, isParent) => {
...
+        fetchedResults = fetchedResults.filter(participant => {
+            return (!queries.firstName || participant.query.firstName.includes(queries.firstName)) &&
+                (!queries.lastName || participant.query.lastName.includes(queries.lastName)) &&
+                (!queries.email || participant.query.allEmails.includes(queries.email)) &&
+                (!queries.phone || participant.query.allPhoneNo.includes(queries.phone)) &&
+                (!queries.dob || participant['371067537'] === queries.dob) &&
+                (!queries.connectId || participant.Connect_ID === parseInt(queries.connectId)) &&
+                (!queries.token || participant.token === queries.token) &&
+                (!queries.studyId || participant.state.studyId === queries.studyId) &&
+                (!queries.checkedIn || participant['331584571.266600170.135591601'] === 353358909);
```
we-ai, line 894: `dob`, `connectId` etc. are common keys used in firstName, lastName and email queries, so they may not need to be checked again after data retrieval.

JoeArmani: I think you are right about this since we build all of these other params into each query.

Example: `firstName` query also contains any other params as compound .where() clauses (dob, connectId, token, studyId, checkedIn). The same is true for `lastName` query and `phone` or `email` query.
So, this filter probably only needs to check `queries.firstName`, `queries.lastName`, `queries.email`, and `queries.phone`.

I need to update this function in the next dev cycle to remove string search support. This will happen once all existing participants are converted to the query.name as array data structures. I will test with the seemingly unnecessary filters removed when I make these updates in the upcoming dev cycle.

PR #384, opened 2023-07-17
handle name search case sensitivity

JoeArmani:
PR: Handle name search case sensitivity.

This is a small update to:
https://github.com/episphere/connectFaas/pull/381
It is in continued support of:
https://github.com/episphere/connect/issues/654

Issue: the query names were being forced to lowercase (desired behavior), but the results filter was filtering based on the original query. This caused lowercase searches to succeed and searches with uppercase characters to fail.

Update: Check for firstName, lastName, and email at the beginning of function execution and convert any/all to lowercase. This improves upon converting the query string to lowercase and then converting the filter to lowercase.

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -811,13 +811,13 @@ const filterDB = async (queries, siteCode, isParent) => {
         for (let key in queryKeys) {
             if (key === 'firstName' || key === 'lastName') {
                 const path = `query.${key}`;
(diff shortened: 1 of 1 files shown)


ramavavi:
The code changes look good.

jhflorey (approved):

we-ai (approved):
Looks good.

PR #385, opened 2023-07-17
Multiple Module Documents - Initial Fix

anthonypetersen:
This PR partially addresses the following Issues:

* https://github.com/episphere/connect/issues/670

Technical Changes

* Updating conditionality in `surveyExists()` from `size === 1` to `size > 0`

----------
Background Details

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -596,7 +596,7 @@ const retrieveUserSurveys = async (uid, concepts) => {
 
 const surveyExists = async (collection, uid) => {
     const snapshot = await db.collection(collection).where('uid', '==', uid).get();
-    if (snapshot.size === 1) {
+    if (snapshot.size > 0) {
         return snapshot.docs[0];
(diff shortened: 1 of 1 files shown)


we-ai (approved):

we-ai:
This should be a result of a subtle bug.

For the comment "I believe if I just close the browser while still logged in, when I go back to stage, it will stay logged in" in [#670](https://github.com/episphere/connect/issues/670), I saw the same error multiple times. It should be a separate bug.

anthonypetersen:
> This should be a result of a subtle bug.
> 
> For the comment "I believe if I just close the browser while still logged in, when I go back to stage, it will stay logged in" in [#670](https://github.com/episphere/connect/issues/670), I saw the same error multiple times in dev site. It should be a separate bug.

I'll create a new issue.

JoeArmani (approved):
looks good

PR #399, opened 2023-08-17
Notifications: Added else if clause for users with both email & phone login method

abhinavjonnada82:
https://github.com/episphere/connect/issues/689

When notifications (esp. emails) are sent out for users with both email & phone login method. The `<loginDetails>` in email is shown as undefined. To prevent this added else if clause to capture users with both login methods.

- Modularized functions that redact email & phone. 
- Returns following statements:
    Email Login: `Your login information for the MyConnect app is ` + `email` + `.`;
    Phone Login: `Your login information for the MyConnect app is` + `phone` + `.`;
    Both Email & Phone Login: `Your login information for the MyConnect app is ` + `email` + `or` + `phone` + `.`;
    
PoC:

Used a workaround to see how the text renders:
  
Phone:
Screenshot 2023-08-17 at 4 02 44 PM

Email:
Screenshot 2023-08-17 at 4 38 23 PM

Both Phone & Email:
Screenshot 2023-08-17 at 4 40 18 PM

### DIFF
--- a/utils/shared.js
+++ b/utils/shared.js
@@ -757,6 +757,18 @@ const isDateTimeFormat = (value) => {
     return typeof value == "string" && (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value));
 }
 
+const redactEmailLoginInfo = (loginDetails) => {
+    let amp = loginDetails.indexOf('@');    
+    for(let i = 0; i < amp; i++) {
+        if(i != 0 && i != 1 && i != amp - 1) {
+            loginDetails = loginDetails.substring(0, i) + "*" + loginDetails.substring(i + 1);
+        } 
+    }
+    return loginDetails
+}
+
+const redactPhoneLoginInfo = (loginDetails) => { return loginDetails = "***-***-" + loginDetails.substring(loginDetails.length - 4); }
+
 module.exports = {
     getResponseJSON,
     setHeaders,
@@ -791,5 +803,7 @@ module.exports = {
     batchLimit,
     getUserProfile,
     isEmpty,
-    isDateTimeFormat
+    isDateTimeFormat,
+    redactEmailLoginInfo,
+    redactPhoneLoginInfo
 }
--- a/utils/shared.js
+++ b/utils/shared.js
@@ -757,17 +757,15 @@ const isDateTimeFormat = (value) => {
     return typeof value == "string" && (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value));
 }
 
-const redactEmailLoginInfo = (loginDetails) => {
-    let amp = loginDetails.indexOf('@');    
-    for(let i = 0; i < amp; i++) {
-        if(i != 0 && i != 1 && i != amp - 1) {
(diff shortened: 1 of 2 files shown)


utils/shared.js
```diff
@@ -757,6 +757,18 @@ const isDateTimeFormat = (value) => {
...
 }
 
+const redactEmailLoginInfo = (loginDetails) => {
+    let amp = loginDetails.indexOf('@');    
+    for(let i = 0; i < amp; i++) {
+        if(i != 0 && i != 1 && i != amp - 1) {
+            loginDetails = loginDetails.substring(0, i) + "*" + loginDetails.substring(i + 1);
+        } 
+    }
+    return loginDetails
```
Gbarra9, line 767: Nice refactor for the email and phone redacting logic. First time seeing a the redacting of characters for email address.Interesting way of redacting the characters before the "@" in the email address.

Gbarra9 (approved):
Looks good

utils/shared.js
JoeArmani, line 767: Looks good, nice way to redact an email address!
Here's an example that also handles a very short email address and avoids rebuilding the string every iteration:
```javascript
const redactEmailLoginInfo = (loginDetails) => {
    const amp = loginDetails.indexOf('@');

    if (amp < 3) return loginDetails; // corner case (first two chars + last char before '@')
    
    let redacted = loginDetails.substring(0, 2); // Keep the first two characters
    
    for (let i = 2; i < amp - 1; i++) {
        redacted += '*';
    }

    redacted += loginDetails.charAt(amp - 1);
    redacted += loginDetails.substring(amp);
    
    return redacted;
};
```

we-ai: I'm doing the same part for speed. And avoided "for" loop:
```js
  const [prefix, domain] = fetchedData[421823980].split("@");
  const changedPrefix = prefix.length > 3
      ? prefix.slice(0, 2) + "*".repeat(prefix.length - 3) + prefix.slice(-1)
      : prefix.slice(0, -1) + "*";
  loginDetails = changedPrefix + "@" + domain;
```
We do have participants whose emails only have 2 letters beside @ and domain.

utils/shared.js
```diff
@@ -757,6 +757,18 @@ const isDateTimeFormat = (value) => {
...
+}
+
+const redactPhoneLoginInfo = (loginDetails) => { return loginDetails = "***-***-" + loginDetails.substring(loginDetails.length - 4); }
```
we-ai, line 770: I used `loginDetails = "***-***-" + fetchedData[348474836].slice(-4);`

we-ai (approved):
Looks good.

utils/shared.js
abhinavjonnada82: Thanks. The corner case is taken care in the original code. I leave it up to @anthonypetersen if he wants to change the function or not.

utils/shared.js
abhinavjonnada82: This is something simpler tho.

utils/shared.js
abhinavjonnada82: https://github.com/episphere/connectFaas/pull/399/files#diff-23fab349744c189c5b655022c3edc921526e74945e34cdd65db82880911119d8R760

we-ai (approved):
Looks good.

JoeArmani (approved):
👍

PR #400, opened 2023-08-21
[#658] Delete data from surveys and speciments

jhflorey:
This PR addresses comment https://github.com/episphere/connect/issues/658#issuecomment-1684110577 from issue https://github.com/episphere/connect/issues/658.

- Added function `removeDocumentFromCollection` to delete all data that relates to participant data destruction from these collections `"bioSurvey_v1", "clinicalBioSurvey_v1", "covid19Survey_v1", "menstrualSurvey_v1", "module1_v1", "module1_v2", "module2_v1", "module2_v2", "module3_v1", "module4_v1", "ssn", "biospecimen"`

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -9,7 +9,13 @@ admin.initializeApp(functions.config().firebase);
 const db = admin.firestore();
 const increment = admin.firestore.FieldValue.increment(1);
 const decrement = admin.firestore.FieldValue.increment(-1);
-const { collectionIdConversion, bagConceptIDs, swapObjKeysAndValues, batchLimit } = require('./shared');
+const {
+    collectionIdConversion,
+    bagConceptIDs,
+    swapObjKeysAndValues,
+    batchLimit,
+    listOfCollectionsRelatedToDataDestruction,
+} = require("./shared");
 const fieldMapping = require('./fieldToConceptIdMapping');
 const { isIsoDate } = require('./validation');
 const nciCode = 13;
@@ -394,23 +400,8 @@ const retrieveParticipantsEligibleForIncentives = async (siteCode, roundType, is
 
 const removeDocumentFromCollection = async (connectID) => {
     try {
-        // Collections with document related to participant
-        const collectionArray = [
-            "bioSurvey_v1",
-            "clinicalBioSurvey_v1",
-            "covid19Survey_v1",
-            "menstrualSurvey_v1",
-            "module1_v1",
-            "module1_v2",
-            "module2_v1",
-            "module2_v2",
-            "module3_v1",
-            "module4_v1",
-            "ssn",
-            "biospecimen",
-        ];
-        while (collectionArray.length > 0) {
-            const collection = collectionArray.shift();
+        while (listOfCollectionsRelatedToDataDestruction.length > 0) {
+            const collection = listOfCollectionsRelatedToDataDestruction.shift();
             const data = await db
                 .collection(collection)
                 .where("Connect_ID", "==", connectID)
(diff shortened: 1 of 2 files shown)


utils/firestore.js
```diff
@@ -392,6 +392,36 @@ const retrieveParticipantsEligibleForIncentives = async (siteCode, roundType, is
...
+    while (collectionArray.length > 0) {
+        const collection = collectionArray.shift();
+        const data = await db
+            .collection(collection)
+            .where("Connect_ID", "==", connectID)
+            .get();
+        if (data.size !== 0) {
+            data.docs.forEach(async (dt) => {
+                await db.collection(collection).doc(dt.id).delete();
+            });
```
JoeArmani, line 420: This is a strange JS nuance that can lead to bugs: `forEach` does not process `async/await` properly and can lead to unexpected behavior.

(for ... of) handles async/await properly.

reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
Quote from reference: `forEach() expects a synchronous function — it does not wait for promises. Make sure you are aware of the implications while using promises (or async functions) as forEach callbacks.`

utils/firestore.js
```diff
@@ -392,6 +392,36 @@ const retrieveParticipantsEligibleForIncentives = async (siteCode, roundType, is
...
+        }
```
JoeArmani, line 421: Will you please wrap this in a try/catch block so we have a way to trace errors?

JoeArmani (commented):
Just comments about forEach() and try/catch. Thanks.

utils/firestore.js
jhflorey: Ok sure  @JoeArmani

utils/firestore.js
jhflorey: @jhflorey  if i use another loop whether it will fix it? Do you have any idea?

JoeArmani: Yes, this does the same thing and works with async/await:
```javascript
for (const dt of data.docs) {
    await db.collection(collection).doc(dt.id).delete();
}
```

jhflorey: Solved

utils/firestore.js
jhflorey: Solved

utils/firestore.js
```diff
@@ -392,6 +392,40 @@ const retrieveParticipantsEligibleForIncentives = async (siteCode, roundType, is
...
+            "clinicalBioSurvey_v1",
+            "covid19Survey_v1",
+            "menstrualSurvey_v1",
+            "module1_v1",
+            "module1_v2",
+            "module2_v1",
+            "module2_v2",
+            "module3_v1",
+            "module4_v1",
+            "ssn",
```
anthonypetersen, line 409: ssn table doesn't contain `Connect_ID`, will never return results

utils/firestore.js
```diff
@@ -392,6 +392,40 @@ const retrieveParticipantsEligibleForIncentives = async (siteCode, roundType, is
     }
 }
 
+const removeDocumentFromCollection = async (connectID) => {
+    try {
+        // Collections with document related to participant
+        const collectionArray = [
```
anthonypetersen, line 398: would prefer we reference an existing constant of survey collection names such as `moduleConceptsToCollections` from `shared.js`

if not, this will just be one more spot in code we need to remember to update when a new survey is added

anthonypetersen (changes requested):

utils/firestore.js
jhflorey: Solved

utils/firestore.js
jhflorey: Solved

jhflorey:
@anthonypetersen Please help re-check again.

anthonypetersen (approved):
looks good

PR #401, opened 2023-08-21
Refactor email handling

we-ai:
This PR solves issue [#692](https://github.com/episphere/connect/issues/692). Below are details:
- After a notification is sent, adds the notification specification IDs to `participants` collection, so that participants fitting specific conditions can be retrieved using only 1 query from BigQuery. This search method is 12+ time faster ( based on tests on dev data) than currently used queries to Firestore.
- To reduce overhead, avoids using new PubSub instances to pass batched queries.
- Send batched emails for speed.
- Use batched data savings to Firestore collections after emails sending.
- Other minor cleanups.


abhinavjonnada82 (approved):

abhinavjonnada82:
Looks good. Nothing to suggest & can move forward.

utils/firestore.js
```diff
@@ -1637,6 +1628,50 @@ const storeNotifications = async payload => {
...
+ * @param {object[]} notificationRecordArray Array of notification records to be saved
+ */
+const saveNotificationBatch = async (notificationRecordArray) => {
+  const chunkSize = 500; // batched write has a doc limit of 500
+  const chunkArray = createChunkArray(notificationRecordArray, chunkSize);
+  for (const recordChunk of chunkArray) {
+    const batch = db.batch();
+    try {
+        for (const record of recordChunk) {
+        const docId = record.uid + record.notificationSpecificationsID.slice(0, 6);
```
anthonypetersen, line 1642: is it necessary to create our own doc id?

utils/notifications.js
```diff
@@ -105,180 +110,184 @@ const sendEmail = async (emailTo, messageSubject, html, cc) => {
...
+  let cutoffTime = new Date();
+  cutoffTime.setDate(cutoffTime.getDate() - notificationSpec.time.day);
+  cutoffTime.setHours(cutoffTime.getHours() - notificationSpec.time.hour);
+  cutoffTime.setMinutes(cutoffTime.getMinutes() - notificationSpec.time.minute);
+  
+  if (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(primaryField)) {
+    // Do nothing if primaryField is a timestamp and time isn't reached. Otherwise, take actions.
+    const scheduledTime = new Date(primaryField);
+    if (scheduledTime > cutoffTime) return;
+    await getParticipantsAndSendEmails({notificationSpec});
```
anthonypetersen, line 139: suggestion: create object prior to first IF statement (i.e. specObject = {notificationSpec})

if you get into ELSE statement, add cutOffTimeStr and timeField to specObject

have one call to getParticipantsAndSendEmails(specObject)

anthonypetersen (approved):
couple of comments, nothing big.. well done, looks good.

utils/firestore.js
we-ai: When using a random ID, each document has a unique ID and is treated as unique, even some documents are identical in contents (eg duplicate emails saved). With a ID generated from uid and spec ID, each notification (to a person) will have only one copy. When a message is saved multiple times, later saves cover earlier saves.

PR #412, opened 2023-08-28
'replaceAll' not supported in node14

we-ai:
This PR fix a subtle bug in PR [#401](https://github.com/episphere/connectFaas/pull/401), by removing 'replaceAll' for string manipulation. 'replaceAll' isn't supported by our current node version 14.


JoeArmani (approved):
√ Use regexp with replace() instead of replaceAll()
√ Remove dotPlaceholder variable and access it directly

looks good

Gbarra9 (approved):
Looks good

anthonypetersen (approved):

PR #415, opened 2023-08-30
backup notifications to BQ

we-ai:
This PR backs up "notifications" collection in Firestore to BigQuery, as described in [#695](https://github.com/episphere/connect/issues/695).
Currently the backup occurs once per day.


we-ai:
> 

I use the data cleanup scheduler which runs once per day.

JoeArmani (approved):
looks good

Question out of curiosity: do we just use the storage bucket for long-term storage as backup data, or are there other uses?

we-ai:
> 

The backups are kept for 1 month, for data recoveries in case any un-intended data changes happen. I may be wrong on this.

anthonypetersen:
> The backups are kept for 1 month, for data recoveries in case any un-intended data changes happen. I may be wrong on this.

Our current rules on the bucket keep backups for 180 days

PR #427, opened 2023-09-13
add getSpecimens: ByCollectionIds & ByReceivedDate

JoeArmani:
Related: https://github.com/episphere/connect/issues/708

Adding endpoints: `getSpecimensByCollectionIds` and `getSpecimensByReceivedDate`

•`getSpecimensByReceivedDate` allows us to query specific boxes and then query only the specimens in those boxes. This will be used in the BPTL dashboard.
•`getSpecimensByCollectionIds` is a helper to `getSpecimensByReceivedDate` and is added as an endpoint for separate specimen search functionality.

PR #428, opened 2023-09-19
update ship API, add getBoxesById()

JoeArmani:
Related: https://github.com/episphere/connect/issues/625
-The ship api has been failing intermittently with no error response. Boxes are not updated as shipped, so the site has to force the shipment with a fake tracking number. This is not ideal or scalable. 

-inspected and updated `ship` endpoint with notes.
-updated `shipBatchBoxes()` function to ensure shipped property is set and run as a Firestore transaction (atomic, ensures data integrity, automatically retries on initial failure).
-added `getBoxesById()` to support get requests on boxes with or without transaction. This will be useful in other scenarios as well.

Note: rules for firestore `'in'` query: max 30 disjunctions. This is preferred because it's sent as a single query. If over 30 disjunctions, we process the request with `Promise.all()`, which is sent as `n` queries.
Disjunctions = (items in array) * (where() clauses).

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -1,4 +1,5 @@
 const { getResponseJSON, setHeaders, logIPAdddress, SSOValidation, convertSiteLoginToNumber } = require('./shared');
+const fieldMapping = require('./fieldToConceptIdMapping');
 
 const biospecimenAPIs = async (req, res) => {
     logIPAdddress(req);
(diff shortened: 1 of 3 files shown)


ramavavi:
Hi Joe, In the code, array of boxIdAndShipmentDataArray into a boxIdToShipmentData object. What is the reason for this change?

JoeArmani:
> Hi Joe, In the code, array of boxIdAndShipmentDataArray into a boxIdToShipmentData object. What is the reason for this change?

Hi @ramavavi, in the endpoint, I was trying to change as little as possible, so I just changed that array from let to const.

In the shipBatchBoxes() function, we updated it to a lookup table by boxId and shipment data for faster lookup and merging during the get/write transaction.

utils/firestore.js
```diff
@@ -1309,35 +1309,83 @@ const updateTempCheckDate = async (institute) => {
...
+            const boxes = await getBoxesByBoxId(boxIdArray, siteCode, false, transaction);
+    
+            for (const box of boxes) {
+                const boxData = box.data;
+                const shipmentData = boxIdToShipmentData[boxData[fieldMapping.shippingBoxId]];
+
+                if (shipmentData) {
+                    shipmentData[fieldMapping.submitShipmentFlag] = fieldMapping.yes;
+                    Object.assign(boxData, shipmentData);
+                    transaction.update(box.docRef, boxData);
```
Gbarra9, line 1377: Initially, I was curious why you were using map to transform each document to `{ data: document.data(), docRef: document.ref }` in the `getBoxesByBoxId` function control flow. This line of code explains why you needed that document.ref, it's for updating. Also, I never used or seen `db.runTransaction`. Thanks for the jsdocs and using it so I can read up on it.

Gbarra9 (approved):
Looks good. No further suggestions.

utils/firestore.js
JoeArmani: Thanks, Gene. Normally, I think the previous batch write was substantial enough for this operation. Since this has been a persistent issue, I like the idea of using a transaction here and tightening up the entire call chain end-to-end. It's a little slower, but we're likely talking about milliseconds. I like that transactions auto-retry and lock access to the associated docs while the transaction runs...it's just an extra layer of certainty that our shipped box data will be consistent.

utils/biospecimen.js
```diff
@@ -359,70 +360,89 @@ const biospecimenAPIs = async (req, res) => {
...
-          145971562: 353358909,
-          105891443: 104430631,
+            ...sharedShipmentData,
+            [fieldMapping.submitShipmentTimestamp]: new Date().toISOString(),
+            [fieldMapping.submitShipmentFlag]: fieldMapping.yes,
+            [fieldMapping.temperatureProbeInBox]: fieldMapping.no,
         };
 
-        let boxIdAndShipmentDataArray = [];
+        const boxIdAndShipmentDataArray = [];
```
we-ai, line 418: I prefer `let` here, to indicate the value may change.

utils/biospecimen.js
```diff
@@ -359,70 +360,89 @@ const biospecimenAPIs = async (req, res) => {
...
 
         for (const boxId of boxIdArray) {
-          let shipmentData = {
-            ...sharedShipmentData,
-            959708259: boxIdToTrackingNumberMap[boxId],
-          };
+            const shipmentData = {
```
we-ai, line 421: I prefer `let` here, to indicate the value may change.

utils/biospecimen.js
```diff
@@ -359,70 +360,89 @@ const biospecimenAPIs = async (req, res) => {
...
+                return res.status(500).json({ message: 'Failed to save box data', code: 500 });
+            }
+            
+            return res.status(200).json({ message: 'Success!', code: 200 });
         } catch (error) {
-          console.log("Error occurred when running shipBatchBoxes():\n", error);
-          return res.status(500).json({ message: 'Internal Server Error', code: 500 });
-        }
-        
+            console.error(`Error occurred when running shipBatchBoxes(): ${error}`);
```
we-ai, line 443:
```suggestion
            console.error(`Error occurred when running shipBatchBoxes(). ${error}`);
```

utils/firestore.js
we-ai: The problem is that, based on my checks on thousands lines of logs, there were no backend errors in those failed shipments: `ship` API was not called. I believe the problem lies in frontend code.

utils/firestore.js
```diff
@@ -1309,35 +1309,83 @@ const updateTempCheckDate = async (institute) => {
...
+        } else if (boxIdArray.length <= 15 && !isBPTL) {
+            const snapshot = await getSnapshot(db.collection('boxes').where(shippingBoxId, 'in', boxIdArray).where(loginSite, '==', siteCode));
+            return snapshot.docs.map(document => ({ data: document.data(), docRef: document.ref }));
+        } else {
+            const queries = boxIdArray.map(boxId => {
+                return db.collection('boxes').where(shippingBoxId, '==', boxId).where(loginSite, '==', siteCode);
+            });
+        
+            const boxResults = await Promise.all(queries.map(query => getSnapshot(query)));
+            return boxResults.map(result => result.docs.map(document => ({ data: document.data(), docRef: document.ref }))).flat();
```
we-ai, line 1344: The 'else' block handles mixed scenarios: BPTL ( when size>30) and non BPTL (when size>15). Please make sure `siteCode` is there for BPTL scenario calls.

we-ai, line 1344: We better have a consistent way to handle different size of data. Same comment here as [there](https://github.com/episphere/connectFaas/pull/427#discussion_r1334779671).

utils/firestore.js
```diff
@@ -1309,35 +1309,83 @@ const updateTempCheckDate = async (institute) => {
 }
 
 /**
- * Ship a batch of boxes
- * @param {Array} boxIdAndShipmentDataArray
- * @param {number} siteCode 
+ * 
+ * @param {Array<string>} boxIdArray - array of box ids to fetch 
+ * @param {string} siteCode - site code of the user (number)
+ * @param {*} isBPTL - boolean to indicate if the request is from BPTL
```
we-ai, line 1315:
```suggestion
 * @param {boolean} isBPTL - boolean to indicate if the request is from BPTL
```

we-ai (approved):
Looks good. Comments on minor things.

utils/biospecimen.js
JoeArmani: Updated to Warren's preference.

utils/biospecimen.js
JoeArmani: Updated to Warren's preference.

utils/firestore.js
JoeArmani: Update: Warren's comment makes me remember this function is useless for BPTL's needs because `boxId` is not unique across sites. I was building this for later use, but it will not be useful. Removed `isBPTL` entirely.

JoeArmani: Update: Warren's suggested pattern is better. Updated to chunking when `boxIdArray.length > 15` and using `Promise.all()` to handle chunked `get()` requests.

utils/firestore.js
JoeArmani: Updated doc (removed isBPTL entirely).

PR #429, opened 2023-09-20
Speed up SMDB dashboard loading

we-ai:
This PR update data retrieval for SMDB dashboard metrics, to solve issue [#23](https://github.com/episphere/dashboard/issues/23). Below are details:
- Added `getStatsFromBQ` to retrieve data from a table based on table name and site code
- Added `getStatsForDashboard` to retrieve all metric data by calling `getStatsFromBQ`
- Other updates:
    * Removed `validateSiteUsers` call in `api === 'validateSiteUsers'` block and return response instead, because `authObj` is always truthy thus `validateSiteUsers` won't run beyond `authObj` check
    * Code cleanup


utils/dashboard.js
```diff
@@ -28,77 +28,71 @@ const dashboard = async (req, res) => {
...
-    else if (api === 'updateParticipantDataNotSite') {
-        if(req.method !== 'POST') {
+    } else if (api === 'updateParticipantDataNotSite') {
+        if (req.method !== 'POST') {
             return res.status(405).json(getResponseJSON('Only POST requests are accepted!', 405));
         }   
         const {submit} = require('./submission');
         let body = req.body;
-        if(!body.uid) {
+        if (!body.uid) {
```
JoeArmani, line 64: I know this is not your code, but there appears to be an error code discrepancy here (line 65).

utils/firestore.js
```diff
@@ -534,27 +534,30 @@ const removeUninvitedParticipants = async () => {
...
         }
-        else{
-            return false;
-        };
+        return [];
     }
     catch(error){
         console.error(error);
-        return new Error(error);
+        return [];
```
JoeArmani, line 560: Question: Does this pattern of not throwing an error still log in DataDog or somewhere on the backend? I'm just asking so I understand how we trace a silent error like this. I do see the conditional handling in `isParentEntity()`...maybe that's my answer?

utils/shared.js
```diff
@@ -479,13 +479,13 @@ const SSOConfig = {
 }
 
 const decodingJWT = (token) => {
-    if(token !== null || token !== undefined){
+    if (token !== null || token !== undefined) {
```
JoeArmani, line 482: Should this be: `if (token !== null && token !== undefined) {`

I haven't traced the JWT token, but it looks like this conditional will always execute.

utils/stats.js
```diff
@@ -59,8 +59,62 @@ const stats = async (req, res, authObj) => {
...
+  let data = {};
+  console.log(`Retrieveing stats data for siteCode: ${siteCodes}`);
+
+  try {
+    let promiseArray = [];
+    for (const shortName of shortNameArray) {
+      promiseArray.push(getStatsFromBQ(shortNameToTableName[shortName], siteCodes));
+    }
+
+    const results = await Promise.all(promiseArray);
```
JoeArmani, line 105: 🥳

JoeArmani (approved):
looks good just left some minor questions/comments.

utils/shared.js
we-ai: I just beautified ( in my point of view) this line, without checking context. I believe it can be shortened to `if (token) {`. Will check further.

utils/firestore.js
we-ai: I prefer this patten, so the returned data have consistent types/shapes. When you re-throw the error, it still need to be handled, which usually means printed, upper in the call stack. Yes, the error message, if occurred in above code, will be printed in backend and all these logs are exported to DataDog. Since we can identify error messages from logs, we may not need to take the efforts to pass error messages further up.

utils/dashboard.js
abhinavjonnada82: Screenshot 2023-09-21 at 2 11 40 PM

utils/stats.js
abhinavjonnada82, line 105: 👍

utils/shared.js
abhinavjonnada82: i would suggest this:`if (token) {`

abhinavjonnada82 (approved):

utils/stats.js
we-ai: > Quick question: Does <Promise.all> handle promise rejections ?

Yes. It rejects when any of the promise rejects. Reference [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all).

utils/dashboard.js
we-ai: Will fix it

utils/dashboard.js
```diff
@@ -28,77 +28,71 @@ const dashboard = async (req, res) => {
...
     const isParent = authObj.isParent;
     const siteCodes = authObj.siteCodes;
     const isCoordinatingCenter = authObj.coordinatingCenter;
     const isHelpDesk = authObj.helpDesk;
     const api = req.query.api;
     console.log(`SMDB API: ${api}, accessed by: ${userEmail}`);
 
-    if(api === 'validateSiteUsers') {
-        const { validateSiteUsers } = require('./validation');
-        return await validateSiteUsers(req, res, authObj);
```
anthonypetersen, line 44: if we truly don't need call to `validateSiteUsers()` anymore, function should be removed as it doesn't appear to be called anywhere else

anthonypetersen (changes requested):
should remove function if not needed anymore

anthonypetersen (approved):
looks good

PR #430, opened 2023-09-20
Biospecimen Dashboard: Update API endpoints to query Shipping Reports

abhinavjonnada82:
Title^^
- Returns Shipping Report data based on filters
- For BPTL, returns reports from all sites
- Optimized code as needed

Related PR:
https://github.com/episphere/biospecimen/pull/585

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1433,12 +1433,12 @@ const getBoxesPagination = async (siteCode, body) => {
     const currPage = body.pageNumber;
     const orderByField = body.orderBy;
     const elementsPerPage = body.elementsPerPage;
-    const filters = body.filters || {};
-    const source = body.source || '';
+    const filters = body.filters ?? {};
+    const source = body.source ?? '';
 
-    const startDate = filters.startDate || '0';
-    const trackingId = filters.trackingId || '';
-    const endDate = filters.endDate || '0';
+    const startDate = filters.startDate ?? '0';
+    const trackingId = filters.trackingId ?? '';
+    const endDate = filters.endDate ?? '0';
 
     let query = db.collection('boxes').where('145971562', '==', 353358909);
 
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1410,136 +1410,63 @@ const getSpecimenCollections = async (token, siteCode) => {
     return [];
 }
 
-const getBoxesPagination = async (siteCode, body) => {
-    let currPage = body.pageNumber;
-    let orderByField = body.orderBy;
-    let elementsPerPage = body.elementsPerPage;
-    let filters = body.filters;
+const buildQueryWithFilters = (query, trackingId, endDate, startDate, source, siteCode) => {
+    if (trackingId !== '') {
+        query = query.where('959708259', '==', trackingId);
+    }
 
-    let startDate = "0";
-    let trackingId = '';
-    let endDate = "0";
+    if (endDate !== '0') {
+        query = query.where('656548982', '<=', endDate);
+    }
 
-    if(filters !== undefined){
-        if(filters.hasOwnProperty('startDate')){
-            startDate = filters['startDate']
-        }
-        if(filters.hasOwnProperty('trackingId')){
-            trackingId = filters['trackingId'];
-        }
-        if(filters.hasOwnProperty('endDate')){
-            endDate = filters['endDate']
-        }
+    if (startDate !== '0') {
+        query = query.where('656548982', '>=', startDate);
     }
-    let snapshot;
-    if(trackingId !== ''){
-        if(endDate !== "0"){
-            if(startDate !== "0"){
-                snapshot =  await db.collection('boxes').where('789843387', '==', siteCode).where('145971562','==',353358909).where('959708259', '==', trackingId).where('656548982', '<=', endDate).where('656548982', '>=', startDate).orderBy(orderByField, 'desc').limit(elementsPerPage).offset(currPage*elementsPerPage).get();
-            }
-            else{
-                snapshot =  await db.collection('boxes').where('789843387', '==', siteCode).where('145971562','==',353358909).where('959708259', '==', trackingId).where('656548982', '<=', endDate).orderBy(orderByField, 'desc').limit(elementsPerPage).offset(currPage*elementsPerPage).get();
-            }
-        }
-        else{
-            if(startDate !== "0"){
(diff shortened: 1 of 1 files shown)


utils/firestore.js
```diff
@@ -1410,136 +1410,63 @@ const getSpecimenCollections = async (token, siteCode) => {
...
+const getBoxesPagination = async (siteCode, body) => {
+    const currPage = body.pageNumber;
+    const orderByField = body.orderBy;
+    const elementsPerPage = body.elementsPerPage;
+    const filters = body.filters ?? {};
+    const source = body.source ?? '';
+
+    const startDate = filters.startDate ?? '0';
+    const trackingId = filters.trackingId ?? '';
+    const endDate = filters.endDate ?? '0';
```
JoeArmani, line 1441: Sugg only: might be cleaner to pass null as the fallback `?? null` and then just check `if (trackingId)` `if (startDate)` `if (endDate)` in `buildQueryWithFilters()`

utils/firestore.js
```diff
@@ -1410,136 +1410,63 @@ const getSpecimenCollections = async (token, siteCode) => {
...
+
+    let query = db.collection('boxes').where('145971562', '==', 353358909);
+
+    query = buildQueryWithFilters(query, trackingId, endDate, startDate, source, siteCode)
+
+    query = query.orderBy(orderByField, 'desc').limit(elementsPerPage).offset(currPage * elementsPerPage);
+
+    const snapshot = await query.get();
```
JoeArmani, line 1449: Seems like it would be good to handle the snapshot error case with `try/catch` here.

utils/firestore.js
```diff
@@ -1410,136 +1410,63 @@ const getSpecimenCollections = async (token, siteCode) => {
...
-    
-}
 
 const getNumBoxesShipped = async (siteCode, body) => {
-    let filters = body;
+    const filters = body;
+    const source = body.source ?? ``;
+    const startDate = filters.startDate ?? '0';
+    const trackingId = filters.trackingId ?? '';
+    const endDate = filters.endDate ?? '0';
```
JoeArmani, line 1461: Sugg only: same comment here about null fallback.

utils/firestore.js
```diff
@@ -1410,136 +1410,63 @@ const getSpecimenCollections = async (token, siteCode) => {
...
-            else{
-                snapshot =  await db.collection('boxes').where('789843387', '==', siteCode).where('145971562','==',353358909).orderBy('656548982', 'desc').get();
-            }
-        }
-    }
+    let query = db.collection('boxes').where('145971562', '==', 353358909);
+
+    query = buildQueryWithFilters(query, trackingId, endDate, startDate, source, siteCode)
+
+    const snapshot = await query.orderBy('656548982', 'desc').get();
```
JoeArmani, line 1467: Question: I'm not sure we need `orderBy()` since we're just returning `.length` but maybe I'm missing something.

JoeArmani (commented):
just minor comments - only important parts are try/catch with snapshots

utils/firestore.js
abhinavjonnada82: I see the purpose of it. Thanks for now, I would prefer not using <null>.

utils/firestore.js
JoeArmani: orderBy() is a great, efficient implementation when sorting is needed.

Might be worth experimenting with .count() in the future. I haven't used this, but it could be very fast compared to pulling data and counting length: Doc: https://cloud.google.com/firestore/docs/query-data/aggregation-queries

JoeArmani (approved):
looks good

utils/firestore.js
```diff
@@ -1410,137 +1410,72 @@ const getSpecimenCollections = async (token, siteCode) => {
...
+
+        query = query.orderBy(orderByField, 'desc').limit(elementsPerPage).offset(currPage * elementsPerPage);
+
+        const snapshot = await query.get();
+        const result = snapshot.docs.map(document => document.data());
+
+        return result;
+    } catch (error) {
+        console.error(error);
+        return new Error(error)
```
we-ai, line 1455: I suggest returning `[]` (same data type as normal return), because this returned error is not handled by calling function.

utils/firestore.js
we-ai: I kind agree with Joe. To make data types consistent, the fallback values can all be "" (empty string), for cleaner condition checks (you can use `if (startDate)` instead of `if (startDate !== '0')`).

utils/firestore.js
```diff
@@ -1410,137 +1410,72 @@ const getSpecimenCollections = async (token, siteCode) => {
...
-        }
+        query = buildQueryWithFilters(query, trackingId, endDate, startDate, source, siteCode)
+
+        const snapshot = await query.orderBy('656548982', 'desc').get();
+        const result = snapshot.docs.length;
+        
+        return result;
+    } catch (error) {
+        console.error(error);
+        return new Error(error)
```
we-ai, line 1477: This returned error is not handled by calling function. Better return `[]`.

utils/firestore.js
```diff
@@ -1410,137 +1410,72 @@ const getSpecimenCollections = async (token, siteCode) => {
...
-                snapshot =  await db.collection('boxes').where('789843387', '==', siteCode).where('145971562','==',353358909).where('656548982', '>=', startDate).orderBy('656548982', 'desc').get();
-            }
-            else{
-                snapshot =  await db.collection('boxes').where('789843387', '==', siteCode).where('145971562','==',353358909).orderBy('656548982', 'desc').get();
-            }
-        }
+        query = buildQueryWithFilters(query, trackingId, endDate, startDate, source, siteCode)
+
+        const snapshot = await query.orderBy('656548982', 'desc').get();
+        const result = snapshot.docs.length;
```
we-ai, line 1472:
```suggestion
        const snapshot = await query.count().get();
        const result = snapshot.data().count;
```

we-ai: As Joe mentioned, there's no need to pull all data (slow and costly), when we need only the count as an integer.

we-ai (approved):
Looks good. Commented on possible further updates.

utils/firestore.js
abhinavjonnada82: Warren I don't think this going to work. 
```
const snapshot = await query.count().get();
const result = snapshot.data().count;
```
this is Web Version 9

PR #431, opened 2023-09-21
Improve bptl's packages in transit manifest loading

Gbarra9:
Related Links: 
[Phase 2 Enhancements](https://github.com/episphere/connect/issues/708) 
[Issue#548](https://github.com/episphere/biospecimen/issues/548)
[Related biospecimen PR#586](https://github.com/episphere/biospecimen/pull/586) 

**This issue addresses the problems of the slow manifest loading page.** 

Initial Page Load:

In the BPTL packages in transit, we currently query all documents from the boxes collection for all sites on the backend for the initial page load. To reduce the loading time, we will pass a req.query.source `bptlPackagesInTransit` to the searchBoxes endpoint. We will query the documents from the boxes collection based on the `submitShipmentFlag (145971562)` being `yes` and `siteShipmentReceived (333524031)` being `no`. This will reduce the documents returned.

On Click Manifest Page Load:

The current client side code passes a single `requestedSite` parameter to query all the `biospecimen` collection documents from the requestedSite value. To reduce loading time we will pass the parameters `requestedSite` and `boxId` and their values to the `searchSpecimen` endpoint. The new function `searchSpecimenBySiteAndBoxId` uses the requestedSite value and `collectionIdArray` (Ex. ['CXA123456', 'CXA212324']) to query for biospecimen documents. This will return the matching documents instead of returning all biospecimen documents. 

**searchSpecimenBySiteAndBoxId function calls:**
1. `searchSpecimenBySiteAndBoxId` calls `getBiospecimenCollectionIdsFromBox` and passes params (requestedSite, boxId) to all chain functions
2. `getBiospecimenCollectionIdsFromBox` function calls `getSiteLocationBox`
3. `getSiteLocationBox` makes a query to get a single box document based on the `requestedSite` value and `boxId` value
4.  `getBiospecimenCollectionIdsFromBox` iterates through the nested box document from `getSiteLocationBox` to extract all the unique collection Ids and stores them into an array (Ex. ['CXA123456', 'CXA212324'])
5. `searchSpecimenBySiteAndBoxId` uses the unique collection Ids from `getBiospecimenCollectionIdsFromBox` and stores the value to a variable collectionIdArray
6. `searchSpecimenBySiteAndBoxId` makes a query on the biospecimen collection to return snapshot of documents with the matching `healthCareProvider` and if the document includes any of the unique collection Ids
7. Iterate through snapshot and push documents to `biospecimenDocs` array
8. return snapshot of documents

**The following was done:**
- modified `searchBoxes` function to accept `bptlPackagesInTransit` to reduce the boxes returned to the intial page load
- added new reference names to fieldmapping such as `loginSite`,`siteShipmentReceived`,`bagContainerCids` obj and more.
- added `else if(req.query.requestedSite && req.query.boxId)` for searchSpecimen endpoint 
- added new function `searchSpecimenBySiteAndBoxId`  when successful returns biospecimen documents matching the collectionId inside the BoxId from a specific site


utils/firestore.js
```diff
@@ -1237,6 +1237,100 @@ const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
...
+ * getBiospecimenCollectionIdsFromBox returns an array of collectionIds from the box
+ * @param {number} requestedSite - site code of the site
+ * @param {string} boxId - boxId of the box
+*/
+const getBiospecimenCollectionIdsFromBox = async (requestedSite, boxId) => {
+    try {
+        const shipBoxMatch = await getSiteLocationBox(requestedSite, boxId);
+
+        if (shipBoxMatch === undefined || shipBoxMatch.length === 0) return [];
+        else {
```
JoeArmani, line 1270: nit: same thing here. Since you're returning in the `if` the `else {}` isn't needed. We can just execute the logic without the `else {}`.

utils/firestore.js
```diff
@@ -1237,6 +1237,100 @@ const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
...
+*/
+const searchSpecimenBySiteAndBoxId = async (requestedSite, boxId) => {
+    try {
+        const collectionIdArray = await getBiospecimenCollectionIdsFromBox(requestedSite, boxId);
+        const snapshot = await db.collection('biospecimen')
+                                .where(fieldMapping.healthCareProvider.toString(), "==", requestedSite)
+                                .where(fieldMapping.collectionId.toString(), "in", collectionIdArray).get();
+        const biospecimenDocs = [];
+
+        if (snapshot.empty) return;
```
JoeArmani, line 1322: nit: I think we want to return `[]` here so the empty array can be passed back as data (just to follow existing patterns)

utils/firestore.js
```diff
@@ -1237,6 +1237,100 @@ const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
...
+        else {
```
JoeArmani, line 1323: nit: since you're returning if (snapshot.empty) the else {} isn't needed. You can just execute the for loop without wrapping it in the else {} clause

utils/firestore.js
```diff
@@ -1369,16 +1463,21 @@ const getLocations = async (institute) => {
...
-    } 
-    else { 
+    if (institute === nciCode || institute == nciConceptId) {
+        if (flag === `bptl`) {
+            snapshot = await db.collection('boxes').get();
+        } else if (flag === `bptlPackagesInTransit`) {
+            snapshot = await db.collection("boxes")
+                            .where(fieldMapping.submitShipmentFlag.toString(), "==", fieldMapping.yes)
+                            .where(fieldMapping.siteShipmentReceived.toString(), "==", fieldMapping.no).get();
+        }
```
JoeArmani, line 1473: nit: for defensiveness we can either include else { (throw error flag is missing) } here OR check for flag in the initial `if` clause. Neither case is expected, but it alerts us if there's a missing flag in one of our API calls coming from BSD:
```javascript
if (flag && (institute === nciCode || institute == nciConceptId)) {
```

JoeArmani (approved):
looks good. Made comments, but minor stuff.

utils/firestore.js
Gbarra9: Thanks for the suggestion.

utils/firestore.js
Gbarra9: Thanks for the suggestion and catching the empty return.

utils/firestore.js
```diff
@@ -1237,6 +1237,100 @@ const searchSpecimen = async (masterSpecimenId, siteCode, allSitesFlag) => {
...
+ * @param {number} requestedSite - site code of the site
+ * @param {string} boxId - boxId of the box
+*/
+const getBiospecimenCollectionIdsFromBox = async (requestedSite, boxId) => {
+    try {
+        const shipBoxMatch = await getSiteLocationBox(requestedSite, boxId);
+
+        if (shipBoxMatch === undefined || shipBoxMatch.length === 0) return [];
+        else {
+            let shipBoxObj = shipBoxMatch[0];
```
jhflorey, line 1271: Do we change value of `shipBoxObj`. If not should we use const instead of let

utils/firestore.js
```diff
@@ -1369,16 +1463,21 @@ const getLocations = async (institute) => {
 
 const searchBoxes = async (institute, flag) => {
     let snapshot = ``
-    if ((institute === nciCode || institute == nciConceptId) && flag === `bptl`) {
-        snapshot = await db.collection('boxes').get()
-    } 
-    else { 
+    if (institute === nciCode || institute == nciConceptId) {
+        if (flag === `bptl`) {
+            snapshot = await db.collection('boxes').get();
```
jhflorey, line 1468: NIT: We cound define `const boxesCollection = db.collection('boxes')`
Then 
`snapshot = await boxesCollection.get()`
`snapshot = await .where(fieldMapping.submitShipmentFlag.toString(), "==", fieldMapping.yes)
                            .where(fieldMapping.siteShipmentReceived.toString(), "==", fieldMapping.no).get();`

ramavavi:
Hi Gene, just had a small query - what is the expected return values from this function : searchSpecimenBySiteAndBoxId?

utils/firestore.js
Gbarra9: I was initially going to add your suggested else block to throw an error, but adding the flag in the initial `if` conditional to check for the existence of the flag parameter makes reading the code easier. Thanks for both suggestions.

utils/firestore.js
Gbarra9: That's a good point. Thanks for catching my incorrect variable declaration. I'll change it to `const` since there is no reassignment.

utils/firestore.js
Gbarra9: Thanks for the feedback. I agree with this change since I am repeating `db.collection('boxes')` twice.

jhflorey:
Looks good to me

jhflorey (approved):

anthonypetersen (approved):
looks good

PR #462, opened 2023-11-15
Streck tube data check

JoeArmani:
Re: intermittent issue with missing streck tube data in prod.

•Add check for streck tube data in firestore biospecimen data object if null, add streck tube placeholder data.
This is now also handled client-side, so we don't expect this check to trigger. It acts as a safety net.


ramavavi (approved):

ramavavi:
Looks good to me.

utils/shared.js
```diff
@@ -1126,6 +1126,35 @@ const sortBoxOnBagRemoval = (boxData, bagsToRemove, currDate) => {
     };
 }
 
+/**
+ * We've had an intermittnt issue with Streck tube data not being added to the specimen data structure (11/2023). 
+ * This is a safety net to ensure the data is added. Build the object literal since we only need to do this for streck tubes.
+ * @param {object} streckTubeData - the streck tube data object (empty placeholder passed in for assignment by reference).
```
jhflorey, line 1132: NIT: it looks like the docblock is missing 1 argument

jhflorey (approved):

utils/shared.js
JoeArmani: nice catch, fixed

utils/shared.js
```diff
@@ -1126,6 +1126,35 @@ const sortBoxOnBagRemoval = (boxData, bagsToRemove, currDate) => {
...
+            [fieldMapping.tubeMislabelledDeviation]: fieldMapping.no,
+            [fieldMapping.tubeDeviationHighTemp]: fieldMapping.no,
+            [fieldMapping.tubeDeviationLowVolume]: fieldMapping.no,
+            [fieldMapping.tubeDeviationLeakedSpilled]: fieldMapping.no,
+            [fieldMapping.tubeDeviationUnexpectedTubeSize]: fieldMapping.no,
+            [fieldMapping.tubeDiscardDeviation]: fieldMapping.no,
+            [fieldMapping.tubeNotFoundDeviation]: fieldMapping.no,
+        },
+    });
+    console.error(`Issue found in updateSpecimen() (ConnectFaas): Streck Tube not found in biospecimenData for collection Id ${collectionId}. Building placeholder data.`);
```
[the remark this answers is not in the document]
JoeArmani: Thanks for the suggestion. In this case, we're actually repairing the data structure above. I just wanted to log the fact that it happened in case there are deeper issues we still need to look into. This same operation is now also handled client-side, so this is more of a sanity check and I don't expect it to trigger (just wanted to make sure the full stack was covered).

anthonypetersen (approved):
solid

PR #491, opened 2023-12-18
Notifications: Update token reference in HP’s self-scheduling link email

abhinavjonnada82:
https://github.com/episphere/connect/issues/846

Function `getParticipantsAndSendEmails()` is only picking up the first TOKEN sign & not the rest. To prevent this, use `replaceAll()` instead `replace` to update multiple references to TOKEN.

`replaceAll()` works for single TOKEN reference as well.

PoC: https://replit.com/@Ajonnada/Nodejs#index.js


Gbarra9 (approved):

bransteitterbr (approved):

JoeArmani (approved):
looks good

PR #494, opened 2023-12-19
Remove replaceAll function from Stage

abhinavjonnada82:
Title^^
 Remove replaceAll function from Stage for issue https://github.com/episphere/connect/issues/846.
 Since we're using older version of JAVASCRIPT.


we-ai:
We're using Node 14. `replaceAll` is available from Node 15, meaning `replaceAll` is not valid in our current Node environment.

we-ai (approved):

anthonypetersen:
is this going to get back ported to dev?

PR #503, opened 2024-01-10
filterParticipantData API

JoeArmani:
Add getFilteredParticipants API

Related issue: https://github.com/episphere/connect/issues/817

-We're splitting off the 'filter' option from the getParticipants API into a separate API.
-The new API can return all participant data or only selected fields specified by the API user.
-The old 'filter' option in the getParticipants API remains in place for now (deprecation warning added)
-Primary in-app use case for the getParticipants 'filter' option is the participant search in SMDB and Biospecimen - this will need to be integrated in the future.


ramavavi (approved):

utils/firestore.js
```diff
@@ -876,23 +880,24 @@ const filterDB = async (queries, siteCode, isParent) => {
...
                 participantQuery = participantQuery.where(path, 'array-contains', queryValue);
             }
-            if (key === 'dob') participantQuery = participantQuery.where('371067537', '==', queries.dob);
-            if (key === 'connectId') participantQuery = participantQuery.where('Connect_ID', '==', parseInt(queries.connectId));
-            if (key === 'token') participantQuery = participantQuery.where('token', '==', queries.token);
-            if (key === 'studyId') participantQuery = participantQuery.where('state.studyId', '==', queries.studyId);
-            if (key === 'checkedIn') participantQuery = participantQuery.where('331584571.266600170.135591601', '==', 353358909);
+            else if (key === 'dob') participantQuery = participantQuery.where('371067537', '==', queries.dob);
+            else if (key === 'connectId') participantQuery = participantQuery.where('Connect_ID', '==', parseInt(queries.connectId));
+            else if (key === 'token') participantQuery = participantQuery.where('token', '==', queries.token);
```
anthonypetersen, line 896: does dashboard code still allow for multiple filter keys to be sent in same call?

utils/submission.js
```diff
@@ -233,18 +233,20 @@ const getParticipants = async (req, res, authObj) => {
         }
     }
     else if(req.query.type === 'filter') {
+        const filterDeprecationWarning = "IMPORTANT: this type == 'filter' API is deprecated. Please use the new 'getFilteredParticipants' API. " + 
+            "Ex: .../getFilteredParticipants?firstName=John&lastName=Doe | API notes & documentation: https://github.com/episphere/connect/issues/817#issuecomment-1883893946";
```
anthonypetersen, line 237: Very nice touch. As a side note, if it hasn't happened already, let's make sure Davin and Michelle know about this so we can get official communication out to the sites.

anthonypetersen (approved):
looks solid.. couple of questions / comments

utils/firestore.js
JoeArmani: It does. I noticed this is an optimization I missed when I rewrote this filterDB function 5 or so months ago. Only one key executes each iteration of the for loop, so the updated if/else if just short-circuits some checks.

I simulated testing with a single site code with isParent=false to emulate Biospecimen participant search
and used an array of site codes with isParent=true to simulate dashboard participant search. Local testing looked good, so I think it's on point.

utils/submission.js
JoeArmani: Thanks, I'll merge and reach out about it now.

PR #517, opened 2024-01-23
fix async bug

we-ai:
This PR fix a bug in calling async function at module level in [PR#516](https://github.com/episphere/connectFaas/pull/516)

### DIFF
--- a/utils/notifications.js
+++ b/utils/notifications.js
@@ -6,8 +6,11 @@ const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, r
 const {getScheduledNotifications, saveNotificationBatch} = require("./firestore");
 const {getParticipantsForNotificationsBQ} = require("./bigquery");
 const conceptIds = require("./fieldToConceptIdMapping");
+let twilioClient, messagingServiceSid;
 
-const [twilioClient, messagingServiceSid] = setupTwilioSms();
+(async () => {
+  [twilioClient, messagingServiceSid] = await setupTwilioSms();
+})();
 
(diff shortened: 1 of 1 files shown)


anthonypetersen (approved):

utils/notifications.js
```diff
@@ -6,8 +6,11 @@ const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, r
 const {getScheduledNotifications, saveNotificationBatch} = require("./firestore");
 const {getParticipantsForNotificationsBQ} = require("./bigquery");
 const conceptIds = require("./fieldToConceptIdMapping");
+let twilioClient, messagingServiceSid;
 
-const [twilioClient, messagingServiceSid] = setupTwilioSms();
+(async () => {
```
bransteitterbr, line 11: This can still lead to a race condition because an async function is still wrapped in a promise and the await inside doesn't block processing.  The safe way to handle this would be to run the setupTwilioSms function with an await in front of it inside the functions that are going to use the twilio client.

we-ai: Good suggestion, and notes taken. I'll make such change next time when updating this repo.

we-ai: Regarding possible race condition, based on long procedures involved in notification sending,  I believe `setupTwilioSms` call will be finished before `twilioClient` is used.

we-ai: I thought about this issue again. In Node microtask queue, `setupTwilioSms` call is way ahead of `sendSmsBatch` call, thus I believe race condition won't occur.

PR #538, opened 2024-02-07
biospecimen reports

JoeArmani:
Related: https://github.com/episphere/connect/issues/755

Optimize reports data
• update searchSpecimenBySiteAndBoxId to handle more collectionIDs (for new big shipping boxes)
• use snapshot.count().get() for reports pagination. This counts docs and returns an integer instead of pulling all data and counting the length. Cost = 1 read per 1000 docs instead of 1 read per doc, data usage drastically reduced.

This merges with: https://github.com/episphere/biospecimen/pull/663

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1436,6 +1436,7 @@ const getBiospecimenCollectionIdsFromBox = async (requestedSite, boxId) => {
  * query the biospecimen collection for documents that match the healthcare provider and collectionIds in collectionIdArray
  * @param {number} requestedSite - site code of the site 
  * @param {string} boxId - boxId of the box
+ * @returns {array} - array of biospecimen documents
  * Firestore 'in' operator limit with two .where() clauses is 15 (30 with one .where() clause).
 */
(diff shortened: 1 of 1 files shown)


utils/firestore.js
```diff
@@ -1436,21 +1436,30 @@ const getBiospecimenCollectionIdsFromBox = async (requestedSite, boxId) => {
  * query the biospecimen collection for documents that match the healthcare provider and collectionIds in collectionIdArray
  * @param {number} requestedSite - site code of the site 
  * @param {string} boxId - boxId of the box
+ * Firestore 'in' operator limit with two .where() clauses is 15 (30 with one .where() clause).
```
jhflorey, line 1439: Should we define the return of this function?

jhflorey (approved):

utils/firestore.js
JoeArmani: Good call - done. Thanks

Gbarra9 (approved):

we-ai (approved):

PR #540, opened 2024-02-10
[#731] Send magic link email with nci email server

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/731

**Code changes**

- Added environment variables for Azure App Registration
- Installed package axios for fetching external services
- Added function ` generateSignInWithEmailLink` from generating the magic link.
- Updated function `getSecrets` for re-using
- Added function `sendEmailLink` for sending email link for participant.

### DIFF
--- a/utils/notifications.js
+++ b/utils/notifications.js
@@ -9,9 +9,9 @@ const conceptIds = require("./fieldToConceptIdMapping");
 
 let twilioClient, messagingServiceSid;
 
-(async () => {
-  [twilioClient, messagingServiceSid] = await setupTwilioSms();
-})();
+// (async () => {
+//   [twilioClient, messagingServiceSid] = await setupTwilioSms();
+// })();
 
 async function setupTwilioSms() {
   const secretsToFetch = {
--- a/utils/notifications.js
+++ b/utils/notifications.js
@@ -2,7 +2,7 @@ const { v4: uuid } = require("uuid");
 const sgMail = require("@sendgrid/mail");
 const showdown = require("showdown");
 const {SecretManagerServiceClient} = require('@google-cloud/secret-manager');
-const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, redactEmailLoginInfo, redactPhoneLoginInfo, createChunkArray, validEmailFormat} = require("./shared");
+const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, redactEmailLoginInfo, redactPhoneLoginInfo, createChunkArray, validEmailFormat, getDatetimeForMagicLink, getEmailTemplateForMagicLink, NIHMailbox} = require("./shared");
 const {getScheduledNotifications, saveNotificationBatch, updateSurveyEligibility} = require("./firestore");
 const {getParticipantsForNotificationsBQ} = require("./bigquery");
 const conceptIds = require("./fieldToConceptIdMapping");
@@ -124,10 +124,10 @@ const retrieveNotifications = async (req, res, uid) => {
     res.status(200).json({data: notifications === false ? [] : notifications, code:200})
 }
 
-const getSecrets = async () => {
+const getSecrets = async (key) => {
     const client = new SecretManagerServiceClient();
     const [version] = await client.accessSecretVersion({
-        name: process.env.GCLOUD_SENDGRID_SECRET,
+        name: key,
     });
     const payload = version.payload.data.toString();
     return payload;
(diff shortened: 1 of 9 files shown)


bransteitterbr:
Is there a reason to use the axios package over using the fetch native library?

jhflorey:
> Is there a reason to use the axios package over using the fetch native library?

I ran into the issue of requiring node-fetch so i have to use axios.

bransteitterbr:
@jhflorey If you are referring to the node-fetch npm package that is no longer required node has fetch built in there is no need to include anything

bransteitterbr:
@jhflorey what version of nodejs are you using? You can check by running `node -v`. connectFass is now using nodejs version 20.  If you haven't yet you should update your local environment and try again.

jhflorey:
@bransteitterbr i've updated to using fetch instead of axios. Thanks for your suggestion.

utils/shared.js
```diff
@@ -1492,6 +1492,39 @@ const filterSelectedFields = (dataObjArray, selectedFieldsArray) => {
...
+      <p><a href="${continueUrl}" target="_other" rel="nofollow">Sign in to Connect for Cancer Prevention Study</a></p>
+      <p>If you did not request this link, you can safely ignore this email.</p>
+      <p>Thanks,</p>
+      <p>Your Connect for Cancer Prevention Study team</p>
+    </body>
+    </html>
+  `;
+};
+
+const NIHMailbox = 'NCIConnectStudy@mail.nih.gov'
```
Gbarra9, line 1527: This is minor and doesn't involve functionality but code styling. At first glance, I was going to suggest renaming `NIHMailbox` to `nihMailbox` to adhere to camelcase. However, I think it's fine since NIH is an acronym.

Gbarra9 (approved):

utils/notifications.js
```diff
@@ -124,18 +124,18 @@ const retrieveNotifications = async (req, res, uid) => {
     res.status(200).json({data: notifications === false ? [] : notifications, code:200})
 }
 
-const getSecrets = async () => {
+const getSecrets = async (key) => {
```
we-ai, line 127: getSecret

utils/notifications.js
```diff
@@ -574,6 +574,85 @@ const getSiteNotification = async (req, res, authObj) => {
...
+                },
+                body: params,
+            }
+        );
+
+        const { access_token } = await resAuthorize.json();
+
+        const body = {
+            message: {
+                subject: `Sign in to Connect for Cancer Prevention Study requested at ${getDatetimeForEmailLink()}`,
```
we-ai, line 616: Do we have a decision about time format in subject? UTC time (eg, 20:43 Z for 3:43pm EST) is confusing.

utils/shared.js
we-ai: I agree that we should use `nihMailbox`.

utils/notifications.js
jhflorey: I implemented it based on old content from firebase authentication.

we-ai: The emails go to our participants, so they better to be clear. Can you confirm that adding time in this format has been discussed and approved by Ops team?

brotzmanmj:
HI @we-ai  No this hasn't been discussed. We should not add the time to the subject line

jhflorey:
Hi @brotzmanmj @we-ai these below image is the subject and content that i tried in https://myconnect.cancer.gov/. Please help confirm our expectation. i'll update it.

image

image

we-ai:
Hi @jhflorey , I believe the only requested change is the link. Nothing else is requested to change. Thus in your snapshots above, the added date/time in email subject and content should be removed, mainly because they cause confusions.
Specifically, subject should be "Sign in to Connect for Cancer Prevention Study".
Email content should be "We received a request to sign in to Connect for Cancer Prevention Study using this email address. If you ...".
Please comment if above info is not accurate @brotzmanmj.

brotzmanmj:
@jhflorey @we-ai  I agree with Warren's comments. @jhflorey do you have an example of what the current Magic Link email looks like? We want it to be exactly the same content for the subject and the body.

jhflorey:
@we-ai @brotzmanmj Please try sign in at https://myconnect.cancer.gov/ to see current email content in prod. i do not add datetime into subject and content.

brotzmanmj:
@jhflorey @we-ai Do NOT sign in to MyConnect in prod. Please always use Stage for any testing.

we-ai:
@brotzmanmj  @jhflorey I mainly use local or dev site for testing. Then do tests in stage to further confirm. I never signed into MyConnect prod site.

we-ai (approved):
Looks good.

PR #542, opened 2024-02-15
Code cleanups and bug fixes

we-ai:
This PR:
- Implements code cleanups described in issue [#871](https://github.com/episphere/connect/issues/871)
- Improves query in BQ
- Improves error handling when retrieving notification schemas

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -1953,21 +1953,15 @@ const retrieveNotificationSchemaByID = async (id) => {
 };
 
 const retrieveNotificationSchemaByCategory = async (category, getDrafts = false) => {
-  let result = [];
-  let query = db.collection("notificationSpecifications");
-  if (category !== "all") query = query.where("category", "==", category);
-  else query = query.orderBy("category");
-
-  // TODO: update query to get daft schemas directly, after all schema in Firestore are updated to have isDraft field
-  const snapshot = await query.orderBy("attempt").get();
-  if (snapshot.size === 0) return result;
-
-  for (const doc of snapshot.docs) {
-    const docData = doc.data();
-    if ((getDrafts && docData.isDraft) || (!getDrafts && !docData.isDraft)) result.push(docData);
+  let query = db.collection("notificationSpecifications").where("isDraft", "==", getDrafts);
+  if (category !== "all") {
+    query = query.where("category", "==", category);
+  } else {
+    query = query.orderBy("category");
   }
 
-  return result;
+  const snapshot = await query.orderBy("attempt").get();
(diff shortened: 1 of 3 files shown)


utils/firestore.js
```diff
@@ -1960,21 +1953,15 @@ const retrieveNotificationSchemaByID = async (id) => {
...
-    if ((getDrafts && docData.isDraft) || (!getDrafts && !docData.isDraft)) result.push(docData);
+  let query = db.collection("notificationSpecifications").where("isDraft", "==", getDrafts);
+  if (category !== "all") {
+    query = query.where("category", "==", category);
+  } else {
+    query = query.orderBy("category");
   }
 
-  return result;
+  const snapshot = await query.orderBy("attempt").get();
```
anthonypetersen, line 1963: will this `orderBy` override the one in the previous `else` statement? or does it act as an additional `orderBy` similar to how a SQL query would work?

utils/notifications.js
```diff
@@ -110,18 +110,26 @@ const markAllNotificationsAsAlreadyRead = (notification, collection) => {
 }
 
 const retrieveNotifications = async (req, res, uid) => {
+  if (req.method !== "GET") {
+    return res.status(405).json(getResponseJSON("Only GET requests are accepted!", 405));
```
anthonypetersen, line 114: unrelated - but thoughts on creating a shared function that can be utilized for common non-200 statuses such as 405 and 500? We could ensure consist messaging and formatting of these responses.

anthonypetersen (approved):
looks good, couple of random comments made that don't affect approval status

utils/firestore.js
we-ai: I believe it works similarly to SQL's way. [Here](https://github.com/firebase/snippets-web/blob/e74e8fd9591f70007a44c58580ff0f1fcd1c7959/firestore/test.firestore.js#L942-L942) is an example used in its documentation.

utils/notifications.js
we-ai: Agree. We should discuss and do it.

anthonypetersen: was a new index required? if so, we will want to remember when code goes to each environment

utils/firestore.js
anthonypetersen: was a new index required? if so, we will want to remember when code goes to each environment

we-ai: Yes. Created dev and stage indexes already. Need your help creating a prod index.

jhflorey (approved):

PR #546, opened 2024-02-22
[#886] Handle missing errorMessage

jhflorey:
This PR addressing issue https://github.com/episphere/connect/issues/1714

The undelivered case does not return the errorMessage, so we have to define it by errorCode based on https://www.twilio.com/docs/messaging/guides/debugging-tools#error-codes

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -4,7 +4,7 @@ admin.initializeApp(functions.config().firebase);
 const db = admin.firestore();
 const increment = admin.firestore.FieldValue.increment(1);
 const decrement = admin.firestore.FieldValue.increment(-1);
-const { tubeConceptIds, collectionIdConversion, swapObjKeysAndValues, batchLimit, listOfCollectionsRelatedToDataDestruction, createChunkArray } = require('./shared');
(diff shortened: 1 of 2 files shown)


Gbarra9 (approved):

utils/firestore.js
```diff
@@ -3010,10 +3009,12 @@ const processTwilioEvent = async (event) => {
             status: event.MessageStatus,
             [`${event.MessageStatus}Date`]: date,
             errorCode: event.ErrorCode || "",
-            errorMessage: event.ErrorMessage || "",
+            errorMessage: event.ErrorMessage || twilioErrorMessages[event.ErrorCode] || "",
         };
 
         await db.collection("notifications").doc(doc.id).update(eventRecord);
+
+        console.log(`Finished updating status for messageSid ${event.MessageSid}:${event.MessageStatus}`)
```
we-ai, line 3017: remove log

we-ai (approved):
Looks good. Minimize printing out logs.

utils/firestore.js
we-ai: No, we don't when there's no errors, because there're so many of messages coming in.
Consider printing out the Sid when error occurs.

PR #549, opened 2024-02-27
Save records timely

we-ai:
This PR is related to issue [#886](https://github.com/episphere/connect/issues/886). It adjusts notification records saving, so that the records can be found and updated based on data from Twilio or SendGrid responses.

### DIFF
--- a/utils/notifications.js
+++ b/utils/notifications.js
@@ -422,12 +422,7 @@ async function getParticipantsAndSendNotifications({ notificationSpec, cutoffTim
           smsCount += smsRecordArray.length;
         }
       } catch (error) {
-        if (error.message.startsWith("saveNotificationBatch")) {
-          console.error(`Error saving SMS records for ${notificationSpec.id}(${readableSpecString}).`, error);
-        } else {
-          console.error(`Error sending SMS messages for ${notificationSpec.id}(${readableSpecString}).`, error);
-        }
-
+        console.error(`Error saving SMS records for ${notificationSpec.id}(${readableSpecString}).`, error);
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


Gbarra9 (approved):

utils/notifications.js
```diff
@@ -401,20 +401,31 @@ async function getParticipantsAndSendNotifications({ notificationSpec, cutoffTim
 
       try {
         await sgMail.send(emailBatch);
+        await saveNotificationBatch(emailRecordArray);
```
JoeArmani, line 404: would Promise.all() be a benefit here or 419 & 421?

utils/notifications.js
```diff
@@ -401,20 +401,31 @@ async function getParticipantsAndSendNotifications({ notificationSpec, cutoffTim
...
+        emailCount += emailRecordArray.length;
       } catch (error) {
-        console.error(`Error sending emails for ${notificationSpec.id}(${readableSpecString}).`, error);
+        if (error.message.startsWith("saveNotificationBatch")) {
+          console.error(`Error saving email records for ${notificationSpec.id}(${readableSpecString}).`, error);
+        } else {
+          console.error(`Error sending emails for ${notificationSpec.id}(${readableSpecString}).`, error);
```
JoeArmani, line 410: Will these be available in DataDog or only in GCP?

JoeArmani (approved):
looks good, only minor questions.

utils/notifications.js
we-ai: GCP logs are imported to Datadog.

utils/notifications.js
we-ai: Since we only save email records after every thing is fine sending out emails, Promise.all() doesn't fit well here. Similar at line 419 and 421.

jhflorey (approved):

PR #555, opened 2024-03-11
PROMIS Survey Scoring

anthonypetersen:
* https://github.com/episphere/connect/issues/870
-----
## Background Details
* Now that the PROMIS Survey has been added to the study we need to add logic to return and store scoring results based on the answers given by each participant
-----
## Technical Changes
* Added `crypto-js` library
* Added file `promis.js`
* `generatePromisAuthToken()` creates authorization token needed for PROMIS API call
* `promisConfig` stores all necessary details for passing results to PROMIS API
* `getScoringData()` actually makes the PROMIS API calls
* `processPromisResults()` grabs participant's survey results and makes the necessary API calls before storing results
* Moved `getSecret()` function from `notifications.js` to `shared.js`
* Added logic condition in `submission.js` to call `processPromisResults()` when PROMIS survey is completed

NOTE: Still need to add API URL for NIH instance of PROMIS, also need to create and reference Secrets for PROMIS uoid and token


amber-emmes (approved):
Looks good.

utils/submission.js
```diff
@@ -700,6 +710,9 @@ const getVersionNumberFromGitHubCommit = async (sha, path, token) => {
     }
 }
 
+// const { processPromisResults } = require('./promis');
+// processPromisResults('eyjekCU9H4V4crXpcRp95HINkXI2');
```
we-ai, line 714: These 2 lines may not be needed.

anthonypetersen: thought I deleted those, thanks for catching

utils/promis.js
```diff
@@ -0,0 +1,578 @@
+const CryptoJS = require('crypto-js');
+
+const generatePromisAuthToken = () => {
+
+    // const { getSecret } = require('./shared');
```
we-ai, line 5: not needed?

anthonypetersen: this will be needed and is a reminder for me... I want to get people testing this in dev, but before this goes to prod we need to change from the testing URL / token to a prod URL / token (which I will keep in a secret)

utils/promis.js
```diff
@@ -0,0 +1,578 @@
...
+                    '367964536': 'Without any difficulty',
+                    '193058048': 'With a little difficulty',
+                    '263412634': 'With some difficulty',
+                    '734992634': 'With much difficulty',
+                    '221192556': 'Unable to do'
+                }
+            }, 
+            'D_917425212': {
+                name: 'PFA21',
+                responses: {
```
we-ai, line 109: Will it be better to save common responses as varaibles and refer to them?

anthonypetersen: Let me try to think how I could minimize this... There were very subtle differences here and there in the responses.

PR #558, opened 2024-03-14
[#636] Add condition to delete participant

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/636#issuecomment-1984579188

ChangeLog

- Add condition to remove participant:  The job removeUninvitedParticipants only removes participants with conditions
    - pin is null
    - token is null
    - verificationStatus(821247024) is not verified(197316935)


amber-emmes (approved):
Looks good

anthonypetersen (approved):

PR #566, opened 2024-03-22
Fix: update uninvited flag not working

jhflorey:
This PR addresses fix issue of https://github.com/episphere/connect/issues/636 

image

### DIFF
--- a/utils/sites.js
+++ b/utils/sites.js
@@ -242,9 +242,15 @@ const updateParticipantData = async (req, res, authObj) => {
         // Reject to update the uninvited flag if the participant is verified or The PIN was used to sign in
         const uninvitedRecruitsId = fieldMapping.participantMap.uninvitedRecruits.toString()
         if (dataObj[uninvitedRecruitsId] && dataObj[uninvitedRecruitsId] === fieldMapping.yes &&
-            (docData[fieldMapping.verified.toString()] === fieldMapping.verified || docData['Connect_ID'])) {
+            (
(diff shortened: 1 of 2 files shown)


utils/sites.js
```diff
@@ -242,9 +242,15 @@ const updateParticipantData = async (req, res, authObj) => {
         // Reject to update the uninvited flag if the participant is verified or The PIN was used to sign in
         const uninvitedRecruitsId = fieldMapping.participantMap.uninvitedRecruits.toString()
         if (dataObj[uninvitedRecruitsId] && dataObj[uninvitedRecruitsId] === fieldMapping.yes &&
-            (docData[fieldMapping.verified.toString()] === fieldMapping.verified || docData['Connect_ID'])) {
+            (
+                docData[fieldMapping.participantMap.signedInFlag.toString()] === fieldMapping.yes
```
amber-emmes, line 246: It looks like you're converting the signedInFlag toString for a triple equality comparison with fieldMapping.yes and fieldMapping.verified, but aren't those constants both numbers? Should probably convert them both to strings for this comparison to be safe.

amber-emmes (changes requested):

utils/sites.js
amber-emmes: Scratch that, misread this.

amber-emmes (approved):

anthonypetersen (approved):

PR #574, opened 2024-04-11
[#959] Set anyRefusalOrWithdrawal to yes

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/959


utils/sites.js
```diff
@@ -264,6 +264,7 @@ const updateParticipantData = async (req, res, authObj) => {
 
         // Handle site notifications
         if(dataObj['831041022'] && dataObj['747006172'] && dataObj['773707518'] && dataObj['831041022'] === 353358909 && dataObj['747006172'] === 353358909 && dataObj['773707518'] === 353358909){ // Data Destruction
+            dataObj['451953807']= 353358909;
```
JoeArmani, line 267: Hi @jhflorey this would mix conditionality that handles site notifications with a participant data update.

While it would work in most cases, I think there's a chance it could fail if the withdrawal or revokeHIPAA variables are set prior to requesting data destruction.

The updateParticipantData API is mixed-use since it handles requests from our apps and from the sites via API. We have a centralized place to check for updates to 'any refusal or withdrawal' 451953807.

Every time `updateParticipantData` is called, the `checkDerivedVariables()` function runs after the participant profile update (see lines 361-362 for that call). Please place this update/check in `checkDerivedVariables()` -> `checkRefusalswithdrawals()` as mentioned here: https://github.com/episphere/connect/issues/959#issuecomment-2047997216

This request is because of the existing structure where we check for refusals/withdrawals and it will help with long-term maintainability. Thanks!

JoeArmani (changes requested):
Hi @jhflorey I'm requesting this update be placed in a different location for consistency with existing code. Please see my comment. Thanks!

PR #591, opened 2024-05-07
Stage -> Dev Sync

anthonypetersen:
Backing some changes needed in stage to dev


JoeArmani (approved):
backing in to sync, already active in stage. looks good

utils/promis.js
```diff
@@ -24,6 +22,7 @@ const processPromisResults = async (uid) => {
 
     const forms = Object.keys(promisConfig);
     const scoresPayload = {};
+    const scoresPromises = [];
```
we-ai, line 25: One promise may not need an array to hold it.

we-ai (approved):

utils/promis.js
anthonypetersen: there are many promises being generated on the `for` loop starting on line 27

anthonypetersen: There are many promises being generated in the

PR #595, opened 2024-05-10
RCA Variables -> add new variables

JoeArmani:
Related: https://github.com/episphere/connect/issues/866

•Add new variables and validation to existing RCA Variables implementation.

### DIFF
--- a/updateParticipantData.json
+++ b/updateParticipantData.json
@@ -535,6 +535,23 @@
 		"dataType": "object",
 		"mustExist": false
 	},
+	"637153953[].844209241": {
+		"values": [104430631, 353358909, 178420302],
+		"dataType": "number",
+		"mustExist": false,
+		"required": true
+	},
+	"637153953[].114227122": {
+		"values": [337516613, 646675764, 178420302],
+		"dataType": "number",
+		"mustExist": false,
+		"required": true
+	},
+	"637153953[].421730068": {
+		"dataType": "string",
+		"mustExist": false,
+		"maxLength": 800
+	},
 	"637153953[].740819233.149205077": {
 		"values": [
 			939782495, 135725957, 518416174, 847945207, 283025574, 942970912, 596122041,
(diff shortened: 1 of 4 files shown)


utils/shared.js
```diff
@@ -1260,6 +1265,22 @@ const validateCancerOccurrence = (cancerSitesObject) => {
     return { error: hasError, message: hasError ? otherCancerSiteErrorMessage : '', data: [] };
 }
 
+/**
+ * Rules: if vitalStatusCategorical is 'alive' at chart review (114227122: 337516613), participant must be aware of diagnosis (844209241: 353358909). Else, block API request.
+ * If vitalStatusCategorical is 'dead' or 'unknown' (114227122: 646675764 or 178420302), participant awareness can be yes, no, or unknown (844209241: 353358909 or 104430631 or 178420302).
+ * @param {number} vitalStatusCategorical - the participant's vital status (conceptID).
+ * @param {number} participantDiagnosisAwareness - the participant's awareness of diagnosis (conceptID).
```
jhflorey, line 1272: NIT: Should we define `return`

jhflorey (approved):

updateParticipantData.json
```diff
@@ -535,6 +535,23 @@
 		"dataType": "object",
 		"mustExist": false
 	},
+	"637153953[].844209241": {
```
we-ai, line 538: Why using `[]` after `637153953`?

JoeArmani: It's for validation of data in a variable length array. The `[]` part enforces the array data structure for validation rules so we can validate any index. Then, during validation, we remove the actual index for an equal comparison.
 
Rules format (JSON): https://github.com/episphere/connectFaas/blob/fa33180d31b0d1249502b295787ca26c80b380b6/updateParticipantData.json#L520
 
Validation (stripping the index to enforce the array structure): https://github.com/episphere/connectFaas/blob/fa33180d31b0d1249502b295787ca26c80b380b6/utils/sites.js#L399

we-ai: I see it now.

utils/shared.js
JoeArmani: Good gatch, thanks. Updated in this function and `validateCancerOccurrence`

we-ai (approved):

PR #601, opened 2024-05-29
Add bswh reported race/ethnicity values for Submit Participant Data and Update Participant Data APIs

Gbarra9:
This PR addresses the following issue:
- https://github.com/episphere/connect/issues/1028

Issue:
The json files for the Submit Participant Data and Update Participant Data API need to have the bswh reported race/ethnicity variable and values.

Code Changes:

- added bswh reported race/ethnicity to submit and update participant jsons


submitParticipantData.json
```diff
@@ -27,6 +27,10 @@
         "values": [628733063, 319570637, 178420302],
         "dataType": "number"
     },
+    "state.253532712": {
+        "values": [723775357, 153444133, 572474909, 308427446, 211228524, 277568192, 611398522, 181769837],
```
bransteitterbr, line 31: should 178420302 be on the list too?

Gbarra9: Thanks for the catch, going to add it now

updateParticipantData.json
```diff
@@ -34,6 +34,10 @@
 		"dataType": "number",
 		"mustExist": true
 	},
+    "state.253532712": {
+        "values": [723775357, 153444133, 572474909, 308427446, 211228524, 277568192, 611398522, 181769837, 178420302],
+        "dataType": "number"
+    },
```
JoeArmani, line 40: Hi @Gbarra9 the "mustExist" property in `updateParticipantData.json` means the property must already exist in the profile to allow it to be updated in the `updateParticipantData` API.

That means the value must have been pushed through the `submitParticipantsData` API initially, and exist in the profile to allow it to be updated. Not sure if you need that here, but I wanted to point out that extra variable you see for other items in the update.json file.

Madhuri will likely know what API is used for this.

`mustExist` usage (applies to the `update` API only, not the `submit` API): https://github.com/episphere/connectFaas/blob/ba0473d59789e008464a17d8c76eacb0f0c44995/utils/sites.js#L426

JoeArmani (approved):
looks good

bransteitterbr (approved):

PR #603, opened 2024-05-31
Spanish and instant notifications

we-ai:
This PR:
- Implements scheduled notifications to handle English or Spanish as preferred language (issue [#918](https://github.com/episphere/connect/issues/918))
- Implements changed instant notification templates location to Firestore (issue [#968](https://github.com/episphere/connect/issues/968))

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -1,5 +1,6 @@
 const { getResponseJSON, setHeaders, logIPAdddress, SSOValidation, convertSiteLoginToNumber } = require('./shared');
 const fieldMapping = require('./fieldToConceptIdMapping');
+const { sendInstantNotification } = require("./notifications");
 
 const biospecimenAPIs = async (req, res) => {
     logIPAdddress(req);
@@ -961,6 +962,21 @@ const biospecimenAPIs = async (req, res) => {
         const response = await sendClientEmail(requestData);
         if(!response) return res.status(404).json(getResponseJSON('ERROR!', 404));
         return res.status(200).json(getResponseJSON('Success!', 200));
+    } else if (api === "sendInstantNotification") {
+        if (req.method !== "POST") {
+          return res.status(405).json(getResponseJSON("Only POST requests are accepted!", 405));
+        }
+
+        const requestData = req.body;
+        try {
+          await sendInstantNotification(requestData);
+          return res.status(200).json(getResponseJSON("Success!", 200));
+        } catch (error) {
+          console.error(
+            `Error sending instant notification (${requestData.category}, ${requestData.attempt}). ${error.message}`
+          );
+          return res.status(500).json({ message: error.message, code: 500 });
+        }
     }
 
     else return res.status(400).json(getResponseJSON('Bad request!', 400));
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -2114,12 +2114,14 @@ const getNotificationSpecById = async (id) => {
 }
 
 const getNotificationSpecByCategoryAndAttempt = async (category = "", attempt = "") => {
+  if (!category || !attempt) return null;
+
   const snapshot = await db
     .collection("notificationSpecifications")
     .where("category", "==", category)
     .where("attempt", "==", attempt)
     .get();
-    
+
   return snapshot.empty ? null : snapshot.docs[0].data();
 };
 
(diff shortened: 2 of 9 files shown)


utils/firestore.js
```diff
@@ -1909,21 +1909,36 @@ const sendClientEmail = async (data) => {
...
     } catch (error) {
         console.error(error);
         return new Error(error);
     }
 }
 
+/**
+ * 
+ * @param {string} userToken User token
+ * @param {string} specId Notification Specifications ID
```
jhflorey, line 1930: NIT: Should we define `return` in DocBlock

utils/firestore.js
```diff
@@ -2375,6 +2400,7 @@ const confirmShipmentKit = async (shipmentData) => {
         const prefEmail = participantDocData['869588347'];
         const token = participantDocData['token'];
         const ptName = participantDocData['153211406'] || participantDocData['399159511']
+        const preferredLanguage = cidToLangMapper[participantDocData[fieldMapping.preferredLanguage]] || "english";
```
jhflorey, line 2403: NIT: Should we define `english` in `shared.js` . We can reuse it somewhere.

jhflorey (approved):

utils/biospecimen.js
```diff
@@ -1,5 +1,6 @@
 const { getResponseJSON, setHeaders, logIPAdddress, SSOValidation, convertSiteLoginToNumber } = require('./shared');
 const fieldMapping = require('./fieldToConceptIdMapping');
+const { sendInstantNotification } = require("./notifications");
```
anthonypetersen, line 3: should we only have this `require` statement inside the `else if` loop where we are using it so that we aren't unnecessarily requiring it when we don't use it?

utils/firestore.js
```diff
@@ -2098,6 +2113,16 @@ const getNotificationSpecById = async (id) => {
     return snapshot.empty ? null : snapshot.docs[0].data();
 }
 
+const getNotificationSpecByCategoryAndAttempt = async (category = "", attempt = "") => {
```
anthonypetersen, line 2116: do we have notifications where the category or attempt would be null? we would be better off null checking them and saving an attempt at searching the database if the parameters are null?

utils/notifications.js
```diff
@@ -2,17 +2,24 @@ const { v4: uuid } = require("uuid");
...
-const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, redactEmailLoginInfo, redactPhoneLoginInfo, createChunkArray, validEmailFormat, getTemplateForEmailLink, nihMailbox, getSecret} = require("./shared");
-const {getNotificationSpecById, getNotificationSpecsByScheduleOncePerDay, saveNotificationBatch, updateSurveyEligibility, generateSignInWithEmailLink} = require("./firestore");
+const {getResponseJSON, setHeadersDomainRestricted, setHeaders, logIPAdddress, redactEmailLoginInfo, redactPhoneLoginInfo, createChunkArray, validEmailFormat, getTemplateForEmailLink, nihMailbox, getSecret, cidToLangMapper} = require("./shared");
+const {getNotificationSpecById, getNotificationSpecByCategoryAndAttempt, getNotificationSpecsByScheduleOncePerDay, saveNotificationBatch, updateSurveyEligibility, generateSignInWithEmailLink, storeNotification, checkIsNotificationSent, getNotificationSpecsBySchedule} = require("./firestore");
 const {getParticipantsForNotificationsBQ} = require("./bigquery");
 const conceptIds = require("./fieldToConceptIdMapping");
 
+const converter = new showdown.Converter();
+const langArray = ["english", "spanish"];
```
anthonypetersen, line 11: better to define langArray as the values from cid lang mapper? if we add more languages in the future, less places we need to update.

utils/notifications.js
```diff
@@ -693,28 +736,30 @@ const sendEmailLink = async (req, res) => {
...
     return res.status(500).json({ data: [], message: JSON.stringify(error, null, 2), code: 500 });
   }
 
-  if (!spec) {
-    message = `Notification spec ID ${req.query.schemaId} isn't found.`;
-    return res.status(404).json({ data: [], message, code: 404 });
-  }
+};
 
+async function handleDruRun(spec) {
```
anthonypetersen, line 762: Dry? *

anthonypetersen (approved):
looks good - few minor comments left

utils/firestore.js
we-ai: Not needed, becuase `english` is defined in `cidToLangMapper` and `fieldMapping` already.
A version without "english" can be: 
```js
const preferredLanguage = cidToLangMapper[participantDocData[fieldMapping.preferredLanguage]] || cidToLangMapper[fieldMapping.english]
```

utils/firestore.js
we-ai: No, we don't have notifications with `null` as category or attempt. So I didn't put those checks.
I agree that we cannot assume the function is correctly used/called, so thorough checking is needed.

utils/biospecimen.js
we-ai: This is something we need to discuss, and maybe some tests are needed to find out which method of import is faster.
My understanding is that if an important is used inside an API, the import is run each time the API is called. For APIs frequently called, putting imports at module level may have performance advantage.

utils/firestore.js
we-ai: The return type is correctly inferred as `Promise<boolean>` by editor, so I didn't repeat its work.

PR #608, opened 2024-06-26
[#1034] Fix: Notifications getting destroyed too soon

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/1034

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -450,7 +450,6 @@ const removeParticipantsDataDestruction = async () => {
         // Check each participant if they are already registered or more than 60 days from the date of their request
         // then the system will delete their data except the stub records and update the dataHasBeenDestroyed flag to yes.
         for (const doc of currSnapshot.docs) {
-            const batch = db.batch();
             const participant = doc.data();
             const timeDiff = isIsoDate(participant[dateRequestedDataDestroyCId])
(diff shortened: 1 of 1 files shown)


bransteitterbr (approved):

utils/firestore.js
```diff
@@ -462,6 +461,7 @@ const removeParticipantsDataDestruction = async () => {
                     requestedAndSignCId ||
                 timeDiff > millisecondsWait
             ) {
+                const batch = db.batch();
```
anthonypetersen, line 464: why is batch being re-initialized for every document?

anthonypetersen (approved):

JoeArmani:
Hi @jhflorey I think this updated batch location could cause an issue with the commit location:
Only one `.commit()` can be called per `batch` created, so please test whether this works when there's more than one iteration in the `for` loop.

I think the `batch.update()` is in the right spot but the `batch.commit()` needs to occur outside the for loop after it finishes iterating. Additional note: this will work for 500 or less operations. Max batch size is 500, so the whole process would need to be handled in a loop if there will be more than 500 items in the for loop.

we-ai:
I agree with Joe. Very likely the `batch.commit()` inside for loop will throw errors.

PR #611, opened 2024-06-27
[#1034] Revert code

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/1034

ChangeLog

- Moved `const batch = db.batch();` inside the condition.

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -451,6 +451,7 @@ const removeParticipantsDataDestruction = async () => {
         // then the system will delete their data except the stub records and update the dataHasBeenDestroyed flag to yes.
         for (const doc of currSnapshot.docs) {
             const participant = doc.data();
+            const participantId = doc.id;
             const timeDiff = isIsoDate(participant[dateRequestedDataDestroyCId])
                 ? new Date().getTime() -
                   new Date(participant[dateRequestedDataDestroyCId]).getTime()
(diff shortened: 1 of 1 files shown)


jhflorey:
@JoeArmani @we-ai @anthonypetersen  Regarding this comment https://github.com/episphere/connectFaas/pull/608#issuecomment-2192446030

- `const batch = db.batch();` was inside the loop before and the PR https://github.com/episphere/connectFaas/pull/608 just only moves it inside the right condition.
- This batch is not used to update multiple participants. It is used to update data of each participant (Remove data + Update 2 participant statuses)

anthonypetersen:
Is there a point in creating batches if we are only updating one participant at a time?

we-ai:
`batch` is meant for batched updates. Using batch to update a single doc is not necessary.

jhflorey:
> `batch` is meant for batched updates. Using batch to update a single doc is not necessary.

Do i need to update to normal way.

anthonypetersen:
I agree with Warren - if you are only updating one document, using `batch` feels unnecessary. You can just use an object to maintain what is going to be updated for the participant and then update the document at the end of your logic.

JoeArmani:
I agree with Tony's and Warren's assessments on this.

jhflorey:
@anthonypetersen @JoeArmani @we-ai just updated and tested with participant (Connect_ID 7009372190)

utils/firestore.js
```diff
@@ -447,12 +447,11 @@ const removeParticipantsDataDestruction = async () => {
             .where(dataHasBeenDestroyed, "!=", fieldMapping.yes)
             .get();
 
-        const batch = db.batch();
-
         // Check each participant if they are already registered or more than 60 days from the date of their request
         // then the system will delete their data except the stub records and update the dataHasBeenDestroyed flag to yes.
         for (const doc of currSnapshot.docs) {
             const participant = doc.data();
+            const participantId = doc.id;
```
anthonypetersen, line 454: NIT - the document ID doesn't represent the ID of the participant

JoeArmani (approved):

utils/firestore.js
we-ai: `docId` is a better name.
This line is not needed, since `doc.id` can be directly used and is clearer.

PR #612, opened 2024-06-27
add getAppSettings endpoint

JoeArmani:
Related: https://github.com/episphere/connect/issues/1048

-add `getAppSettings` to fetch Firestore `appSettings` properties
-add ConnectApp endpoint

NOTE: Merge with ConnectApp https://github.com/episphere/connectApp/pull/801

### DIFF
--- a/utils/connectApp.js
+++ b/utils/connectApp.js
@@ -73,6 +73,7 @@ const connectApp = async (req, res) => {
       
       return res.status(200).json({data: shaResult, code: 200});
     }
+
     else if (api === 'getSHAFromGitHubCommitData') {
       if (req.method !== 'GET') {
         return res.status(405).json(getResponseJSON('Only GET requests are accepted!', 405));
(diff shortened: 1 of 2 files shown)


bransteitterbr (approved):

anthonypetersen (approved):

utils/connectApp.js
```diff
@@ -91,6 +92,25 @@ const connectApp = async (req, res) => {
       return res.status(200).json({data: shaResult, code: 200});
     }
 
+    else if (api === 'getAppSettings') {
+      if (req.method !== 'GET') {
+        return res.status(405).json(getResponseJSON('Only GET requests are accepted!', 405));
+      }
+
+      const selectedParamsArray = req.query.selectedParamsArray
```
bransteitterbr, line 100: This is more of a long term design consideration then a comment on this specific code. From a security standpoint as we design apis we should consider why things liked the selected parameters array is optional.  Are there a lot of use cases where the caller is going to need all the data?  If not, then requiring the caller to specify the data they need is both more secure and in the long term more cost effective and performant because it will reduce payload sizes and in turn lower egress costs.

JoeArmani: @bransteitterbr Great point - agree and updated so `selectedParamsArray` is required. Thanks https://github.com/episphere/connectFaas/pull/612/commits/5d5344818bc205ad193d8f1dfab4ed87baa86cb3

PR #615, opened 2024-07-02
temp endpoint

anthonypetersen:
* N/A
-----
## Background Details
* Undiscovered bug when PROMIS survey was released prevented scoring API from being called when participants completed the survey
* Need to manually trigger these results to get processed
* Due to API being behind firewall we need a temporary end-point that can access the API when called from a local source
-----
## Technical Changes
*


JoeArmani (approved):
looks good

PR #617, opened 2024-07-08
Duplicate tracking numbers now prohibited (1038)

amber-emmes:
Validation work to prevent duplicate tracking numbers, per [1038](https://github.com/episphere/connect/issues/1038). Requires corresponding biospecimen work for full functionality.

### DIFF
--- a/utils/biospecimen.js
+++ b/utils/biospecimen.js
@@ -671,14 +671,14 @@ const biospecimenAPIs = async (req, res) => {
         if( req.method !== 'GET') {
             return res.status(405).json(getResponseJSON('Only GET requests are accepted!', 405));
         }
-        const supplyQuery = req.query.supplyKitId;
-        const collectionQuery = (req.query.collectionId?.slice(0, -4) || "") + " " + (req.query.collectionId?.slice(-4) || ""); // add space to collection
-        const returnKitTrackingNumberQuery = req.query.returnKitTrackingNumber;
+        const supplyKitId = req.query.supplyKitId;
+        const collectionIdSuffix = (req.query.collectionId?.slice(0, -4) || "") + " " + (req.query.collectionId?.slice(-4) || ""); // add space to collection
+        const returnKitTrackingNumberNumber = req.query.returnKitTrackingNumber;
         if (Object.keys(query).length === 0) return res.status(404).json(getResponseJSON('Please include id to check uniqueness.', 400));
-        if (collectionQuery.length < 14) return res.status(200).json({data: 'Check Collection ID', code:200});
+        if (collectionIdSuffix.length < 14) return res.status(200).json({data: 'Check Collection ID', code:200});
         try {
             const { checkCollectionUniqueness } = require('./firestore');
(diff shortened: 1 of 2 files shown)


bransteitterbr (approved):

utils/biospecimen.js
```diff
@@ -673,11 +673,12 @@ const biospecimenAPIs = async (req, res) => {
         }
         const supplyQuery = req.query.supplyKitId;
         const collectionQuery = (req.query.collectionId?.slice(0, -4) || "") + " " + (req.query.collectionId?.slice(-4) || ""); // add space to collection
+        const returnKitTrackingNumberQuery = req.query.returnKitTrackingNumber;
```
we-ai, line 676: It's a tracking number, not a query, right?

amber-emmes: It's the tracking number pulled from the query. I'm following the naming convention from supplyQuery and collectionQuery, but I'm happy to change it if you'd rather.

utils/firestore.js
```diff
@@ -2152,16 +2152,29 @@ const updateKitAssemblyData = async (data) => {
...
 
-const checkCollectionUniqueness = async (supplyId, collectionId) => {
+const checkCollectionUniqueness = async (supplyId, collectionId, returnKitTrackingNumber) => {
     try {
         const supplySnapShot = await db.collection('kitAssembly').where('690210658', '==', supplyId).get();
         const collectionSnapShot = await db.collection('kitAssembly').where('259846815', '==', collectionId).get();
-        if (supplySnapShot.docs.length === 0 && collectionSnapShot.docs.length === 0) {
+        let returnKitTrackingNumberSnapshot = {docs: []};
+        let supplyKitTrackingNumberSnapshot = {docs: []};
+        if(returnKitTrackingNumber) {
```
we-ai, line 2161: The `returnKitTrackingNumber` can be checked earlier.

utils/firestore.js
```diff
@@ -2295,10 +2308,39 @@ const processParticipantHomeMouthwashKitData = (record, printLabel) => {
...
+            .get();
+
+        if(kitsWithDuplicateReturnTrackingNumbers.size > 0) {
+            return false;
+        }
+
+        const otherKitsUsingSupplyKitTrackingNumber = await db.collection("kitAssembly")
+            .where(`${supplyKitTrackingNum}`, '==', data[supplyKitTrackingNum])
+            .get();
+            // .where(`${supplyKitId}`, '!=', data[supplyKitId]).get();
```
we-ai, line 2330: Remove this line if not needed.

we-ai (approved):

utils/biospecimen.js
we-ai: I don't think those names should be followed. They should be updated instead for clarity.

utils/firestore.js
we-ai: Assuming bad value or missing values of `returnKitTrackingNumber` are passed from API calls, it can be a little more performant to do checks at API entry point. But current way works.

PR #625, opened 2024-07-20
[#1057] Type conversation

jhflorey:
This PR addresses issue https://github.com/episphere/connect/issues/1057

### DIFF
--- a/utils/notifications.js
+++ b/utils/notifications.js
@@ -705,8 +705,7 @@ const sendEmailLink = async (req, res) => {
         const body = {
             message: {
                 subject:
-                    preferredLanguage ===
-                    conceptIds.spanish.toString()
+                    preferredLanguage === conceptIds.spanish
                         ? "Inicie sesión para Estudio Connect para la Prevención del Cáncer"
                         : "Sign in to Connect for Cancer Prevention Study",
                 body: {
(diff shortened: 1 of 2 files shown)


utils/shared.js
```diff
@@ -1546,7 +1546,7 @@ const getTemplateForEmailLink = (
     continueUrl,
     preferredLanguage = fieldMapping.english.toString()
 ) => {
-    return preferredLanguage === fieldMapping.spanish.toString()
+    return preferredLanguage.toString() === fieldMapping.spanish.toString()
```
we-ai, line 1549: No need to convert to strings if they both are numbers.

jhflorey: Solved

Gbarra9 (approved):

utils/notifications.js
```diff
@@ -705,7 +705,7 @@ const sendEmailLink = async (req, res) => {
         const body = {
             message: {
                 subject:
-                    preferredLanguage ===
+                    preferredLanguage.toString() ===
```
we-ai, line 708: No need to convert to strings if they both are numbers.

jhflorey: Solved

we-ai (approved):

PR #642, opened 2024-09-09
Kit assembly now enforces uppercase and checks for matching ID number…

amber-emmes:
…s (1089)

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -2298,8 +2298,32 @@ const getNotificationSpecByCategoryAndAttempt = async (category = "", attempt = 
   return snapshot.empty ? null : snapshot.docs[0].data();
 };
 
+const validateKitAssemblyData = (data) => {
+    // Ensure that values are uppercase and that the appropriate values match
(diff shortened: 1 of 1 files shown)


utils/firestore.js
```diff
@@ -2298,8 +2298,32 @@ const getNotificationSpecByCategoryAndAttempt = async (category = "", attempt =
   return snapshot.empty ? null : snapshot.docs[0].data();
 };
 
+const validateKitAssemblyData = (data) => {
+    // Ensure that values are uppercase and that the appropriate values match
+    if(data[fieldMapping.returnKitId]) {
+        data[fieldMapping.returnKitId] = ('' + data[fieldMapping.returnKitId]).toUpperCase();
```
anthonypetersen, line 2304: What is the purpose of adding the empty strings in all these statements?

amber-emmes: I do it to make 100% sure it's a string before running toUpperCase, in case we end up with a number.

anthonypetersen (approved):

Gbarra9 (approved):

PR #646, opened 2024-09-16
Block MC reminder after 45 days

we-ai:
Temporarily add hard-coded 45-day restriction on MC survey reminders, to fix prod issue.
A long-term solution will be implemented shortly in September release.


JoeArmani (approved):
looks good

PR #654, opened 2024-09-20
Stop time and upgraded spec conditions handling

we-ai:
Details in this PR:
- Support stop time for notifications, as requested in [issue#1112](https://github.com/episphere/connect/issues/1112)
- Upgrade handling of notification specification conditions, as requested in [issue#1064](https://github.com/episphere/connect/issues/1064)
- Some cleanups

PR #655, opened 2024-09-20
Handle edge case

we-ai:
Update for [PR#654](https://github.com/episphere/connectFaas/pull/654), to handle possible OR conditions without parentheses.


amber-emmes (approved):
Looks good.

PR #667, opened 2024-10-03
Fix error code 12300 in Twilio

we-ai:
Fix error code 12300 in Twilio, as it doesn't prefer responses with JSON data.


jhflorey (approved):

JoeArmani (approved):
Just curiosity...we now log the error but everything returns 204? Could the JSON issue and 12300 error be related to Twilio’s TwiML messaging format? https://stackoverflow.com/questions/66133189/twilio-function-to-call-external-api-invalid-content-type-error-code-12300

we-ai:
Tested in dev: error 12300 is fixed in Twilio with the changed responses.

PR #668, opened 2024-10-08
Updated error handling in instant notifications

we-ai:
Details of this PR:
- Print more detailed info when an error occurs when sending instant notifications, for easier troubleshooting
- Send status code 204 to Twilio after handling incoming messages, so that it doesn't complain with error 11200


amber-emmes (approved):
Looks good.

PR #670, opened 2024-10-17
norc birthday cards implementation

JoeArmani:
Related: https://github.com/episphere/connect/issues/888

NORC Birthday Cards - return data

Handled:
•return data write to birthdayCard collection.
•'create' case: new data.
•'update' case: updated data (such as changed dispCode and/or added returnDate).
•integration with user profile updates and history.
•integration with .query profile arrays (for search functions).

Working Spec: https://nih.app.box.com/file/1570377597275

Note: We expect additional 'dispCode' options from NORC, and I'll put in a quick PR for those updates when we receive them.


bransteitterbr (approved):

anthonypetersen (approved):
looks solid

PR #688, opened 2024-11-18
Initial home mouthwash work, re-implemented and tested.

amber-emmes:


utils/validation.js
```diff
@@ -181,6 +181,42 @@ const getToken = async (req, res) => {
...
+        (
+            !data[conceptIds.collectionDetails][conceptIds.baseline][conceptIds.bioKitMouthwash] ||
+            data[conceptIds.collectionDetails][conceptIds.baseline][conceptIds.bioKitMouthwash][conceptIds.kitStatus]
+        )
+    ) {
+        const isEligible = !!processParticipantHomeMouthwashKitData(data, true);
+        if(isEligible) {
+            updates[`${conceptIds.collectionDetails}.${conceptIds.baseline}.${conceptIds.bioKitMouthwash}.${conceptIds.kitStatus}`] = conceptIds.initialized;
+        }
+    } else if(
```
JoeArmani, line 205: Would converting this to an `if` block (instead of `else if`)and using the updates object:
`updates[conceptIds.collectionDetails][conceptIds.baseline][conceptIds.bioKitMouthwash]`
eliminate the need for the second `processMouthwashEligibility` call (outside the transaction) in `submitSpecimen`?

I'm just curious when this `else if` case is applicable. When is kitStatus is set to initialized and it needs to be reverted?

JoeArmani (changes requested):
Left some questions and comments.

Also, did anything prompt the changes to `package-lock.json`? If not, please reset those changes.

utils/validation.js
amber-emmes: The else if case is in case the participant was previously eligible but updated to be ineligible due to changing their address to a PO box or something. Somewhat of an edge case but I figured it was best to include it.

utils/firestore.js
```diff
@@ -1300,6 +1300,109 @@ const storeSpecimen = async (data) => {
...
+        const {checkDerivedVariables} = require('./validation');
+        await checkDerivedVariables(participantToken, siteCode);
+        // Now process mouthwash eligibility
+        // This needs the above to run and so is outside of the transaction
+        // First get the updated participant data
+        const participantSnapshot = await db.collection('participants')
+            .where('state.uid', '==', participantUid)
+            .get();
+        const eligibilityUpdates = processMouthwashEligibility(participantSnapshot.docs[0].data());
+        console.log('Outer eligibilityUpdates', JSON.stringify(eligibilityUpdates, null, '\t'));
```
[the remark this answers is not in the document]
amber-emmes: It's not actually necessary; I just missed removing the previous call when I had to add it outside of the transaction. Removing it.

amber-emmes:
Comments addressed.

JoeArmani:
Looking good, how about this?
 
> Also, did anything prompt the changes to `package-lock.json`? If not, please reset those changes.

bransteitterbr (approved):

PR #692, opened 2024-11-26
await response from 'submit' endpoint

JoeArmani:
Related: https://github.com/episphere/connect/issues/1168 

Update the submit endpoint to await the response and return a response for all cases.

Implementation: This is used for response confirmation, in-app handling & communication, and question reversion when a response to a question fails to update server-side in Quest 2.

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -101,16 +101,19 @@ const updateResponse = async (data, uid) => {
     try{
         const snapshot = await db.collection('participants').where('state.uid', '==', uid).get();
         printDocsCount(snapshot, "updateResponse");
-        if(snapshot.size === 1) {
-            for(let doc of snapshot.docs){
-                await db.collection('participants').doc(doc.id).update(data);
-                return true;
-            }
+
+        if (snapshot.size !== 1) {
+            throw new Error(`updateResponse expected 1 document, found ${snapshot.size}. uid: ${uid}`);
         }
+
+        const docId = snapshot.docs[0].id;
+        
+        await db.collection('participants').doc(docId).update(data);
+        return true;
     }
     catch(error){
-        console.error(error);
-        return new Error(error)
+        console.error(`Error in updateResponse: ${error}`);
+        return new Error(`Error in updateResponse: ${error}`);
     }
 }
 
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -106,9 +106,7 @@ const updateResponse = async (data, uid) => {
             throw new Error(`updateResponse expected 1 document, found ${snapshot.size}. uid: ${uid}`);
         }
 
-        const docId = snapshot.docs[0].id;
-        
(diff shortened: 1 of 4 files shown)


utils/firestore.js
```diff
@@ -101,16 +101,19 @@ const updateResponse = async (data, uid) => {
...
-                return true;
-            }
+
+        if (snapshot.size !== 1) {
+            throw new Error(`updateResponse expected 1 document, found ${snapshot.size}. uid: ${uid}`);
         }
+
+        const docId = snapshot.docs[0].id;
+        
+        await db.collection('participants').doc(docId).update(data);
```
we-ai, line 111:
```suggestion
        await snapshot.docs[0].ref.update(data);
```

utils/shared.js
```diff
@@ -784,7 +784,7 @@ const cleanSurveyData = (data) => {
     const admin = require('firebase-admin');
     
     Object.keys(data).forEach(key => {
-        if(data[key] === null) {
+        if(data[key] == null) {
```
we-ai, line 787: Should we use strick check here `data[key] === null || data[key] === undefined`?

utils/submission.js
```diff
@@ -2,95 +2,99 @@ const { getResponseJSON, setHeaders, logIPAddress } = require('./shared');
...
-        keys.forEach(key => {
-            const { moduleStatusConcepts } = require('./shared');
+        if (response) {
+            let moduleComplete = false;
+            let calculateScores = false;
 
-            if (moduleStatusConcepts[key] && data[key] === 231311385) {
-                moduleComplete = true;
+            keys.forEach(key => {
+                const { moduleStatusConcepts } = require('./shared');
```
we-ai, line 56: `moduleStatusConcepts` can be cached, or import is done at top of file.

utils/submission.js
```diff
@@ -2,95 +2,99 @@ const { getResponseJSON, setHeaders, logIPAddress } = require('./shared');
...
-            const ID = generateConnectID();
-            const response = await sanityCheckConnectID(ID);
-            if(response === true) {
-                Connect_ID = ID;
-                boo = true;
+    try {
+        // generate Connect_ID if Consent form submitted
+        if(data[919254129] !== undefined && data[919254129] === 353358909) {
+            const { generateConnectID } = require('./shared');
+            const { sanityCheckConnectID } = require('./firestore');
```
we-ai, line 13: I feel that multiple rounds of imports from `shared.js` and `firestore.js` slows down code. It could be faster if imports are done at top of file.

we-ai (approved):
Looks good. Thanks for the cleanups.
Comments on minor things.

utils/firestore.js
JoeArmani: Thanks, updated

utils/shared.js
JoeArmani: sounds good, updated

PR #720, opened 2025-01-16
Testing + workflow updates

amber-emmes:
Includes the following changes:

1. Home mouthwash kit consideration now includes physical address. ([1174](https://github.com/episphere/connect/issues/1174))
2. Biospecimen submission workflow will no longer overwrite data when dealing with data from multiple finalized collections ([1062](https://github.com/episphere/connect/issues/1062))
3. Fixes an issue with a false positive in the else case for validating home mouthwash eligibility
4. Mocha testing preliminary implementation: adds Mocha unit test cases for a number of functions, including some additional dev-only packages necessary for tests and a few additional exports used by test cases.


jhflorey (approved):

test/test.js
```diff
@@ -1,80 +1,1948 @@
-const assert = require('assert');
+// import {assert} from 'chai';
```
we-ai, line 1: not needed

package.json
```diff
@@ -36,7 +36,10 @@
   },
   "devDependencies": {
     "@google-cloud/functions-framework": "^3.2.0",
+    "chai": "^4.5.0",
     "mocha": "^10.2.0",
+    "node-mocks-http": "^1.16.0",
+    "sinon": "^19.0.2",
     "supertest": "^6.3.3"
```
we-ai, line 43: Should "mocha" and "supertest" be upgraded?

we-ai (approved):
Looks good. Minor comments.

package.json
amber-emmes: As I recall, I kept these at their current versions due to breaking incompatibilities with some other packages we weren't yet ready to upgrade to the newest version.

PR #721, opened 2025-01-17
Update first sign in timestamp and sign-up handling

JoeArmani:
Related: https://github.com/episphere/connect/issues/986

•Update Sign-up handling (First sign-in time, PIN, and related variables).
•Change GET reqs to new POST reqs since those operations write data.
•Retain backward compatibility with old endpoints while MyConnect code can be cached on participant devices.


Gbarra9 (approved):

we-ai (approved):
Looks great! I don't have anything to add to the updates you've done.
Besides the updates, should we be proactive and wait for async function executions in `connectApp.js`?  These functions are not awaited: sendEmailLink, submitSocial, getUserSurveys, retrieveNotifications, validateUsersEmailPhone

JoeArmani:
> Looks great! I don't have anything to add to the updates you've done. Besides the updates, should we be proactive and wait for async function executions in `connectApp.js`? These functions are not awaited: sendEmailLink, submitSocial, getUserSurveys, retrieveNotifications, validateUsersEmailPhone

Thanks @we-ai, I like that suggestion. I took time to test all those with `await` this morning because I was concerned about hidden bugs. Everything appears to work well.

PR #725, opened 2025-01-23
Add a route to get the physical activity report

bransteitterbr:
The backend portion of the ROI Physical Activity Report https://github.com/episphere/connect/issues/1163

### DIFF
--- a/utils/reports.js
+++ b/utils/reports.js
@@ -1,0 +1,44 @@
+const { getBigQueryData } = require('./bigquery');
+const { retrieveConnectID } = require('./firestore');
+const { getResponseJSON } = require('./shared');
+
+const retrievePhysicalActivityReport = async (req, res, uid, connectId) => {
+    if (req.method !== "GET") {
+        return res.status(405).json(getResponseJSON("Only GET requests are accepted!", 405));
+    }
+
+    //If there is a uid passed in lookup the connect id from the participant
+    if (uid) {
+        connectId = await retrieveConnectID(uid);
+    }
+
+    //If the connectId  an object (error from retrieveConnectID)
+    if (typeof connectId === 'object') {
+        return res.status(500).json(getResponseJSON("Error Finding Connect ID", 500));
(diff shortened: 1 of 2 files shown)


amber-emmes (approved):
Looks good.

utils/reports.js
```diff
@@ -1,4 +1,43 @@
...
+const { retrieveConnectID } = require('./firestore');
+const { getResponseJSON, logIPAddress, setHeaders } = require('./shared');
+
+const retrievePhysicalActivityReport = async (req, res, uid, connectId) => {
+    if (req.method !== "GET") {
+        return res.status(405).json(getResponseJSON("Only GET requests are accepted!", 405));
+    }
+
+    //If there is a uid passed in lookup the connect id from the participant
+    if (uid) {
```
anthonypetersen, line 11: error handling if `uid` isn't present

utils/reports.js
```diff
@@ -1,4 +1,43 @@
...
+    //If the connectId  an object (error from retrieveConnectID)
+    if (typeof connectId === 'object') {
+        return res.status(500).json(getResponseJSON("Error Finding Connect ID", 500));
+    }
+    //If the connectId is missing then we can not continue
+    else if (!connectId) {
+        return res.status(400).json(getResponseJSON("Connect ID Not Passed", 400));
+    }
+
+    connectId = '6759772253';
```
anthonypetersen, line 24: assuming this was for testing? if so, please remove

anthonypetersen (changes requested):
see comments

utils/reports.js
bransteitterbr: This function is going to be used by both connectApp and smdb and in the case of the smdb it will pass in the connectId.  If the uid  is missing or the connectId is not passed or it fails to lookup then the error handling on lines 16 or 20 will catch it

anthonypetersen (approved):

PR #749, opened 2025-03-10
Update p.o. box filter for assign kits

Gbarra9:
The PR is related to:
- https://github.com/episphere/connect/issues/1248

Original filter does not handle spacing after "p" for po box.

Code Change:
- added optional spacing after the P for regex matching


jhflorey (approved):

JoeArmani (approved):
looks good

PR #751, opened 2025-03-12
Replacement home mw kits

amber-emmes:
Adds API support for replacement home MW kits (part of [1142](https://github.com/episphere/connect/issues/1142))


utils/firestore.js
```diff
@@ -2757,24 +2832,159 @@ const queryKitsByReceivedDate = async (receivedDateTimestamp) => {
...
-        const { addressPrinted, collectionDetails, baseline, bioKitMouthwash, bloodOrUrineCollectedTimestamp, kitStatus } = fieldMapping;
+        const { addressPrinted, collectionDetails, baseline, bioKitMouthwash, bioKitMouthwashBL1, bioKitMouthwashBL2, bloodOrUrineCollectedTimestamp, dateKitRequested, kitStatus } = fieldMapping;
 
-        const snapshot = await db.collection("participants")
-            .where(`${collectionDetails}.${baseline}.${bioKitMouthwash}.${kitStatus}`, '==', addressPrinted)
-            .select(...participantHomeCollectionKitFields)
-            .orderBy(`${collectionDetails}.${baseline}.${bloodOrUrineCollectedTimestamp}`, 'desc')
-            .get();
-        printDocsCount(snapshot, "eligibleParticipantsForKitAssignment");
+        const participantSnapshotPromises = [
```
JoeArmani, line 2837: Please either implement or remove the commented-out orderBy query segments.

Just mentioning this might be a good candidate for an `OR` query.

No need to make changes, just nice to know about: https://firebase.google.com/docs/firestore/query-data/queries
Note: Unfortunately, I don't think orderBy() is compatible with OR queries.

utils/firestore.js
```diff
@@ -2757,24 +2832,159 @@ const queryKitsByReceivedDate = async (receivedDateTimestamp) => {
...
+            console.error('Error getting participant replacement kit', err);
+            throw err;
+        }
+    });
+}
+
+/**
+ * Broken into a separate helper function for easier testing
+ */
+const getHomeMWReplacementKitData = (data) => {
```
JoeArmani, line 2919: Please consider placing helper functions that don't access firestore outside of the firestore.js file

utils/firestore.js
```diff
@@ -2811,9 +3021,53 @@ const addKitStatusToParticipant = async (participantsCID) => {
...
+                case 'BL_1':
+                    path = bioKitMouthwashBL1;
+                    break;
+                case 'BL_2':
+                    path = bioKitMouthwashBL2;
+                    break;
+                default:
+                    path = bioKitMouthwash;
+            }
+            const snapshot = await db.collection("participants").where('Connect_ID', '==', parseInt(participantCID)).select('id').get();
```
JoeArmani, line 3042: Is 'id' a valid property in .select('id')?
Should that be .select('Connect_ID')?

utils/firestore.js
```diff
@@ -2811,9 +3021,53 @@ const addKitStatusToParticipant = async (participantsCID) => {
...
+            printDocsCount(snapshot, "addKitStatusToParticipantV2");
+            if (snapshot.size === 0) {
+                // No matching document found, stop the update
+                return false;
+            }
+            const docId = snapshot.docs[0].id;
+            await db.collection("participants").doc(docId).update({
```
JoeArmani, line 3049: I think this `await` actually slows down the function and makes it execute sequentially instead of concurrently.

We're mapping all the promises and then awaiting them later in Promise.all(), so please test removal of this await.

utils/firestore.js
```diff
@@ -3083,14 +3356,25 @@ const confirmShipmentKit = async (shipmentData) => {
...
-            const participantQuery = db.collection("participants")
+
+            // Needs multiple participant queries for all cases
+            const participantQueryInitial = db.collection("participants")
                 .where(`${collectionDetails}.${baseline}.${bioKitMouthwash}.${uniqueKitID}`, '==', shipmentData[uniqueKitID]);
-            const participantSnapshot = await transaction.get(participantQuery);
+            const participantQuery1 = db.collection("participants")
+                .where(`${collectionDetails}.${baseline}.${bioKitMouthwashBL1}.${uniqueKitID}`, '==', shipmentData[uniqueKitID]);
+            const participantQuery2 = db.collection("participants")
+                .where(`${collectionDetails}.${baseline}.${bioKitMouthwashBL2}.${uniqueKitID}`, '==', shipmentData[uniqueKitID]);
```
JoeArmani, line 3370: Possibly another good candidate for a Firestore `OR` query.

JoeArmani (commented):
left comments, some minor things and some informational only.

utils/validation.js
```diff
@@ -212,16 +212,25 @@ const processMouthwashEligibility = (data) => {
...
+            if(data?.[conceptIds.collectionDetails]?.[conceptIds.baseline]?.[conceptIds.bioKitMouthwash]?.[conceptIds.kitStatus] == conceptIds.initialized) {
+                updates[`${conceptIds.collectionDetails}.${conceptIds.baseline}.${conceptIds.bioKitMouthwash}.${conceptIds.kitStatus}`] = undefined;
+            }
+            // Handle replacement kits as well
+            if(data?.[conceptIds.collectionDetails]?.[conceptIds.baseline]?.[conceptIds.bioKitMouthwashBL1]?.[conceptIds.kitStatus] == conceptIds.initialized) {
+                updates[`${conceptIds.collectionDetails}.${conceptIds.baseline}.${conceptIds.bioKitMouthwashBL1}.${conceptIds.kitStatus}`] = undefined;
+            }
+            if(data?.[conceptIds.collectionDetails]?.[conceptIds.baseline]?.[conceptIds.bioKitMouthwashBL2]?.[conceptIds.kitStatus] == conceptIds.initialized) {
+                updates[`${conceptIds.collectionDetails}.${conceptIds.baseline}.${conceptIds.bioKitMouthwashBL2}.${conceptIds.kitStatus}`] = undefined;
+            }
```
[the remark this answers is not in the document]
amber-emmes: Correct.

utils/firestore.js
amber-emmes: It's to get only the Firestore-generated ID for use in .doc later. AFAIK this is the correct syntax, confirmed by testing.

JoeArmani: I think `.select('id')` would look for a field we stored named 'id' (similar to 'Connect_ID' and 'state.uid').

Then, this .select() call would return an empty data body. The fetch process would still acknowledge the document exists and provide access to the autogenerated document id property (as part of the document's metadata), giving you access to the document id with snapshot.docs[0].id.

If that's the case, I think it's less confusing for future teammates to use .select('Connect_ID') to keep the return lightweight but use properties that exist in our data structures.

JoeArmani (approved):
looks good

utils/shared.js
```diff
@@ -1039,6 +1039,149 @@ const updateBaselineData = (biospecimenData, participantData, siteTubesList) =>
...
+    }
+    
+   const participantIsEligible = !!processParticipantHomeMouthwashKitData(data, true);
+
+   if(!participantIsEligible) {
+    throw new Error('Participant address information is invalid.');
+   }
+
+   
+    if(data?.[fieldMapping.collectionDetails]?.[fieldMapping.baseline]?.[fieldMapping.bioKitMouthwashBL1]) {
```
anthonypetersen, line 1059: NIT - assigning some of these super long field paths to variables might help readibility

utils/firestore.js
```diff
@@ -3195,23 +3343,20 @@ const storeKitReceipt = async (pkg) => {
             transaction.set(newDocRef, biospecPkg);
 
             transaction.update(kitDoc.ref, {
-                '137401245': pkg['137401245'] === true ? 353358909 : 104430631,
-                '221592017': 375535639,
-                '633640710': processPackageConditions(pkg['633640710']),
-                '755095663': pkg['755095663'],
-                '826941471': pkg['826941471']
+                [collectionCardFlag]: pkg[collectionCardFlag] === true ? yes : no,
```
anthonypetersen, line 3346: thank you for replacing these with variables

anthonypetersen (approved):
looks good - small comments left

PR #754, opened 2025-03-21
Commented out functionality not ready for stage

amber-emmes:


bransteitterbr (approved):

we-ai (commented):
What issue does this change tie to?
What's the plan adding it back?
Do we need to modify March release summary because of this removed functionality?

amber-emmes:
> What issue does this change tie to? What's the plan adding it back? Do we need to modify March release summary because of this removed functionality?

It ties into the replacement kit logic work. It will be added back in when everything else with the replacement kit work is released to stage and prod. We do not need to modify the March release summary due to it.

PR #759, opened 2025-03-28
Get Participants API Refactor

anthonypetersen:
Removing `offset` functionality from `getParticipants` API

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -579,7 +579,7 @@ const retrieveParticipants = async (siteCode, type, isParent, limit, cursor, fro
 
             if (cursor) {
                 const collection = 'participants';
-                const doc = await getCursorDocumnet(collection, cursor);
+                const doc = await getCursorDocument(collection, cursor);
 
                 if (doc.exists) {
                     query = query.startAfter(doc);
@@ -619,7 +619,7 @@ const retrieveParticipants = async (siteCode, type, isParent, limit, cursor, fro
  * Retrieves a document from a Firestore collection using its ID for pagination purposes.
  *
  * @async
- * @function getCursorDocumnet
+ * @function getCursorDocument
  * @param {string} collection - The name of the Firestore collection to query.
  * @param {string} cursor - The document ID to retrieve, typically the last document ID from a previous query result.
  * @returns {Promise<FirebaseFirestore.DocumentSnapshot|Error>} A Promise that resolves to:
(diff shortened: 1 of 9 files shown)


jhflorey (approved):

utils/firestore.js
```diff
@@ -530,7 +578,8 @@ const retrieveParticipantsDemo = async (siteCode, type, isParent, limit, cursor,
             }
 
             if (cursor) {
-                const doc = await getCursorDocumnet('participants', cursor);
+                const collection = 'participants';
+                const doc = await getCursorDocument(collection, cursor);
```
JoeArmani, line 582: I think doc may either exist or be an `instanceof Error` here if the cursor isn't found.

utils/firestore.js
```diff
@@ -486,31 +486,79 @@ const resetParticipantHelper = async (uid, saveToDb) => {
...
+            },
+            'refusalswithdrawals': {
+                where: [[refusalConcept, '==', 353358909]],
+                orderBy: ['Connect_ID', 'asc']
             }
-        }
+        }  
 
-        const applyConditions = async (query, type, siteCode, limit, cursor, from, to) => {
+        const applyConditions = async (collection, type, siteCode, limit, cursor, from, to) => {
```
JoeArmani, line 559: I think Firestore is specific about inequality operators `'>='` `'<='` and `.orderby()` clauses, so when an inequality operator is used, the first `orderby()` needs to match the term in the inequality operator.  It looks like there could be a case where the `orderby` attribute in the conditions obj is appended before the fromTo orderby, which coincides with the inequality operator (lines 571-578). This might be a non-issue and Firestore may handle it automatically. Just wanted to mention it in case.

JoeArmani (approved):
Looks great! This will have a huge impact. Left two minor questions/comments.

utils/firestore.js
anthonypetersen: I'll test this out!

utils/firestore.js
anthonypetersen: can you clarify what you mean here?

JoeArmani: My mistake...the else branch handles it. I'm wondering if we should return that new Error from the else block so the query doesn't execute?

utils/firestore.js
anthonypetersen: tested out cases with inequality clauses, Firestore seems to hand it appropriately!

PR #765, opened 2025-04-17
Eligible participants sort updated per request

amber-emmes:
* Fixed issue where .data() was not getting called
* Sort order now from oldest to newest rather than newest to oldest
* Sort for replacement kits now uses date kit requested rather than bloodOrUrineCollectedTimestamp
* Addresses with PO boxes which have made it this far are no longer filtered out


JoeArmani (approved):
looks good

jhflorey (approved):

PR #773, opened 2025-04-28
Updated the collection stats queries and processing

bransteitterbr:
Modified the queries to use _JP versions of the tables because the original tables only exist in dev.  For issue: https://github.com/episphere/connect/issues/1041


amber-emmes (approved):
Looks good; approved.

JoeArmani (approved):
looks good

PR #778, opened 2025-05-08
Further YAML Separation

anthonypetersen:
* N/A
-----
## Technical Changes
* Splitting Cloud Functions into their own YAML files to speed up Cloud Builds

### DIFF
--- a/config/dev/biospecimen.yaml
+++ b/config/dev/biospecimen.yaml
@@ -1,11 +1,10 @@
 steps:
 
-# getParticipants
 - name: 'gcr.io/cloud-builders/gcloud'
   args: [
     'functions', 
     'deploy', 
-    'getParticipants', 
+    '${_FUNCTION}', 
     '--trigger-http', 
     '--runtime=${_RUNTIME}', 
     '--memory=${_MEMORY}',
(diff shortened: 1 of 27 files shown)


anthonypetersen:
> Looks good. What dictates which functions get their own yaml files with build information? Is this something to be considered for future workflow purposes?

Good question. 

I've wanted to do this for a while as some Cloud Builds were taking a long time. Physical Activity got its own initially as it was a new function around the time I decided to do this. Others got them as I was already updating the code within them. Then I wanted to split our front-end apps into their own as well as grouping others that made sense (incentives, site participant updates, etc.)

Moving forward, if a new Cloud Function is added I would prefer that a new YAML file is added. I'm hoping the template I've set up makes it easy to do.

we-ai:
Looks good.
With increased number of yaml files, we may need a common pattern of file names (connectFaas-something or myconnect-something ?), to avoid confusions since there're build triggers set up by analytics team.

we-ai (approved):

anthonypetersen:
> Looks good. With increased number of yaml files, we may need a common pattern of file names (connectFaas-something or myconnect-something ?), to avoid confusions since there're build triggers set up by analytics team.

I will work on a common naming structure

PR #779, opened 2025-05-13
Fixes error handling for add and update kit data to be consistent with expectations and handling in biospecimen. 

amber-emmes:
Fixes error handling for add and update kit data to be consistent with expectations and handling in biospecimen. This change was made as a part of the cleanup work for [1133](https://github.com/episphere/connect/issues/1133) after realizing that several errors were not being returned as expected from the API. Changes:

firestore.js:
* Errors from addKitAssemblyData and updateKitAssemblyData are now thrown and not returned, preventing messages from being swallowed

biospecimen.js
* error cases in corresponding endpoints to above now return errors with explicit correct object formatting and account for possible lack of error.message

Tests done:
* Automatic tests re-run
* Tested reused kit IDs and tracking numbers on kitAssembly page and ensured that errors displayed correctly.

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -2503,42 +2503,32 @@ const validateKitAssemblyData = async (data) => {
 }
 
 const addKitAssemblyData = async (data) => {
-    try {
-        await validateKitAssemblyData(data);
-        await db.collection('kitAssembly').add(data);
-        
-        return true;
-    }
-    catch(error){
-        console.error(error);
-        return new Error(error);
-    }
+    // No try/catch; that is handled in the biospecimen endpoint
+    await validateKitAssemblyData(data);
+    await db.collection('kitAssembly').add(data);
+    
+    return true;
 }
 
 
 const updateKitAssemblyData = async (data) => {
-    try {
-        await validateKitAssemblyData(data);
-        const snapshot = await db.collection('kitAssembly').where('687158491', '==', data['687158491']).get();
-        printDocsCount(snapshot, "updateKitAssemblyData");
+    // No try/catch; that is handled in the biospecimen endpoint
+    await validateKitAssemblyData(data);
+    const snapshot = await db.collection('kitAssembly').where('687158491', '==', data['687158491']).get();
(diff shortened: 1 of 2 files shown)


JoeArmani (approved):
looks good.

It would be helpful to swap out the hardcoded concept IDs in `updateKitAssemblyData` for fieldMapping variables for long-term maintainability.

utils/firestore.js
```diff
@@ -2503,42 +2503,32 @@ const validateKitAssemblyData = async (data) => {
...
 }
 
 
 const updateKitAssemblyData = async (data) => {
-    try {
-        await validateKitAssemblyData(data);
-        const snapshot = await db.collection('kitAssembly').where('687158491', '==', data['687158491']).get();
-        printDocsCount(snapshot, "updateKitAssemblyData");
+    // No try/catch; that is handled in the biospecimen endpoint
+    await validateKitAssemblyData(data);
```
Gbarra9, line 2516: https://github.com/NCI-C4CP/connectFaas/blob/7ff02f15310690aa1486a99f9ae01feec95658cb/utils/firestore.js#L2477-L2503

I had to revisit this function. As you mentioned from the PR, is the try catch block here clearing out the possible "throw error message" from `validateKitAssemblyData` function because we are returning a new error object? By removing this try catch here we can get the correct error message in both `updateKitData` and `addKitData `?

Gbarra9 (approved):
Looks good

amber-emmes:
> looks good.
> 
> It would be helpful to swap out the hardcoded concept IDs in `updateKitAssemblyData` for fieldMapping variables for long-term maintainability.

I'm inclined to agree, but I'm leery of doing it in this release, since it would require re-testing the function to make sure everything ends up the correct keys still vs. a more cursory test of functionality and error handling.

utils/firestore.js
amber-emmes: Correct. Also, because it was returning an error instead of throwing it, it was not being caught by the try/catch in the old code and was resulting in all responses having 200 codes and the response not being the expected string.

PR #784, opened 2025-05-23
Map DHQ survey status (internal) to DHQ survey status (external)

JoeArmani:
Map DHQ survey status (internal) to DHQ survey status (external)

Related: https://github.com/episphere/connect/issues/1238

We're using a different set of Concept IDs to refer to DHQ survey status found in the DHQ back end. This updates survey status mapping to use those variables.


Gbarra9 (approved):
Looks good

PR #795, opened 2025-06-16
Upload pathology reports to SMDB

we-ai:
This PR implement backend of issue [#1244](https://github.com/episphere/connect/issues/1244).  Below are details: 
- Create buckets based on participant's health care provider
- Save uploaded files to designated folder of participants, in buckets
- Save file names in "pathologyReports" collection in Firestore

### DIFF
--- a/utils/fileUploads.js
+++ b/utils/fileUploads.js
@@ -6,6 +6,17 @@ const storage = new Storage();
 const tierStr = process.env.GCLOUD_PROJECT.split("-").slice(3, 5).join("-").toLowerCase();
 const pathologyReports = "pathology-reports";
 
+async function awaitNewBucketReady(bucketName, maxRetries = 10, delayMs = 500) {
+  let retries = 0;
+  while (retries < maxRetries) {
+    const [exists] = await storage.bucket(bucketName).exists();
+    if (exists) return;
+    await new Promise((res) => setTimeout(res, delayMs));
+    retries++;
+  }
+  throw new Error(`Bucket ${bucketName} not ready after ${maxRetries} retries`);
+}
+
 const uploadPathologyReports = (req, res) => {
   if (req.method !== "POST") {
     return res.status(405).json({ message: "Only POST requests are accepted!", code: 405 });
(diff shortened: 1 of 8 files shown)


utils/fileUploads.js
```diff
@@ -0,0 +1,134 @@
...
+        stream.on("error", reject);
+      })
+    );
+  });
+
+  busboy.on("finish", async () => {
+    try {
+      const promiseResults = await Promise.allSettled(uploadPromises);
+      const successFilenames = promiseResults
+        .filter((result) => result.status === "fulfilled")
```
amber-emmes, line 79: This seems like it will swallow errors if some of the files upload but some of them don't. Should this be recorded somewhere?

amber-emmes (approved):
Looks good. One note, not blocking.

utils/fileUploads.js
```diff
@@ -0,0 +1,133 @@
...
+        .map((result) => result.value);
+
+      if (successFilenames.length === 0) {
+        return res.status(500).json({ message: "Failed to upload files!", code: 500 });
+      }
+
+      await savePathologyReportNamesToFirestore({ bucketName, Connect_ID: parseInt(Connect_ID), filenames: successFilenames });
+
+      const [allFiles] = await bucket.getFiles({ prefix: `${Connect_ID}/` });
```
JoeArmani, line 91: Can we use `successFilenames` here instead of fetching with `getFiles`?

utils/fileUploads.js
JoeArmani: Is there a straightforward way to compare the initial file list with `successFilenames` so we know which ones failed in the event of service issues and or possibly return/log `206 partial` like we do in the API when some items fail? Or does this mix current and previously updated data?

JoeArmani (approved):
Looks good! just minor comments

utils/fileUploads.js
we-ai: `successFilenames` contains names of files just uploaded.
And `getFiles`  retrieves all files currently exist in the directory.
We try to display all filenames under the directory to frontend, so `getFiles` is more suitable here.

utils/fileUploads.js
we-ai: It's easy to get failed filenames and return them, if errors occur. 206 might be the most appropriate code when mixed cases happen.
I'll do it shortly.

PR #798, opened 2025-06-18
Clean-up: Rm updateDHQReportViewedStatus from dashboard.js

JoeArmani:
Related: https://github.com/episphere/connect/issues/1327

Rm updateDHQReportViewedStatusendpoint from dashboard.js. We will only update the report viewed status from participant actions in the PWA, so we don't use this endpoint in dashboard.js.


amber-emmes (approved):
Looks good.

anthonypetersen (approved):

PR #800, opened 2025-06-18
Home MW Kit Updates

amber-emmes:
Per Erin's and Michelle's testing and feedback on [1020](https://github.com/episphere/connect/issues/1020), the following changes were made to firestore.js:

storeKitReceipt:
Updated kit receipt not to overwrite research MW collection data not nested in a specific kit path per request. This means that if a research MW collection is done before a home MW kit is receipted, the data will continue to reflect the research MW collection.

eligibleParticipantsForKitAssignment:
Fixed an issue with sorting on the kit assignment page. Data was not being passed in in the appropriate format to the sort function in this location, leading to data displaying in the wrong order on the kit assignment page.

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -2829,9 +2829,19 @@ const eligibleParticipantsForKitAssignment = async () => {
             }
         })
 
-        manualRequests = manualRequests.sort(manualRequestSort);
+        manualRequests = manualRequests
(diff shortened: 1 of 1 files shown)


Gbarra9 (approved):
looks good

JoeArmani (approved):
Looks good

PR #810, opened 2025-07-02
Per feedback, R2 home MW kits can now have addressUndeliverable updated

amber-emmes:
Changes:
test.js:
* New test case added for undeliverable R2 kit scenario
* Some test description updates for accuracy

shared.js:
* getHomeMWKitData updated so that when passed a participant whose R2 kit status is marked as address undeliverable, it will update the R2 kit to initialized.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1468,7 +1468,7 @@ describe('biospecimen', async () => {
                 });
     
             });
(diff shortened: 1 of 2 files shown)


bransteitterbr (approved):

bransteitterbr:
Looks Good

JoeArmani (approved):
Looks good

PR #812, opened 2025-07-09
Delete pathology reports and Firestore records

we-ai:
This PR implements requirements in [issue#1353](https://github.com/episphere/connect/issues/1353), for data destruction daily cleanups. Details below:
- Delete pathology reports in bucket
- Delete docs in "pathologyReports" collection in Firestore

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -797,17 +797,28 @@ const deletePathologyReports = async (connectId) => {
 
     await Promise.all(fileDeletePromises);
 
+    let batch = db.batch();
     let counter = 0;
-    const batch = db.batch();
+    const batchLimit = 500;
+    const batchPromises = [];
     for (const doc of snapshot.docs) {
       const { [fileNameCidStr]: fileName, bucketName } = doc.data();
       if (failedDeletionSet.has(`${bucketName}/${connectId}/${fileName}`)) continue;
       batch.delete(doc.ref);
       counter++;
+      if (counter >= batchLimit) {
+        batchPromises.push(batch.commit());
+        batch = db.batch();
+        counter = 0;
+      }
     }
 
     if (counter > 0) {
-      await batch.commit();
+      batchPromises.push(batch.commit());
+    }
+
+    if (batchPromises.length > 0) {
+      await Promise.all(batchPromises);
     }
   } catch (error) {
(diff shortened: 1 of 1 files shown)


utils/firestore.js
```diff
@@ -745,6 +745,75 @@ const removeDocumentFromCollection = async (connectID, token) => {
     }
 };
 
+/**
+ * Deletes pathology report files from Storage Bucket and file records from "pathologyReports" collection in Firestore
+ * @param {number} connectId
+ * @returns Promise<void>
+ */
+const deletePathologyReports = async (connectId) => {
+  const tierStr = process.env.GCLOUD_PROJECT?.split("-").slice(3, 5).join("-").toLowerCase() || "";
```
anthonypetersen, line 754: might be easier / cleaner to just create an environment variable in the config yaml file?

utils/firestore.js
```diff
@@ -745,6 +745,75 @@ const removeDocumentFromCollection = async (connectID, token) => {
...
+    }
+
+    await Promise.all(fileDeletePromises);
+
+    let counter = 0;
+    const batch = db.batch();
+    for (const doc of snapshot.docs) {
+      const { [fileNameCidStr]: fileName, bucketName } = doc.data();
+      if (failedDeletionSet.has(`${bucketName}/${connectId}/${fileName}`)) continue;
+      batch.delete(doc.ref);
```
anthonypetersen, line 805: do we feel we need to have any checks that batch size doesn't go over the 500(?) limit?

anthonypetersen (approved):
few questions / comments, very clean and looks good

utils/firestore.js
we-ai: The `GCLOUD_PROJECT` defined inside `.env.yaml` is good enough. We may not need to create environment variables for special cases like this.
For bucket names, leads suggested to use shorter (`connect-dev` for dev tier) instead of longer string (`nih-nci-dceg-connect-dev` or `dceg-connect-dev` for dev tier). Thus the truncated project name is used here to find buckets.

utils/firestore.js
we-ai: Agreed that we should handle the limit, although the possibility is low that some participants have more than 500 pathology reports.

bransteitterbr (approved):

PR #813, opened 2025-07-09
1370 updates

amber-emmes:
Updates different stages of the home MW kit process to appropriately handle deceased and withdrawn users at each point in time. For [1370](https://github.com/episphere/connect/issues/1370).

Changes:

biospecimen.js
* confirmShipment: Error changed to include error message as response (previous behavior returns error message under key message instead of response, leading to error being swallowed on front end)

site.js
* updateParticipantData now clears kit status for participants with home MW kits in the status of initialized or address undeliverable.

submission.js
* Cleared some commented out unused code for readability

firestore.js
* assignKitToParticipant now checks for deceased and withdrawn participants, clears their relevant kit status, and informs the user
* markParticipantAddress undeliverable now checks for deceased and withdrawn participants, clears their relevant kit status, and informs the user
* confirmShipmentKit format changed slightly to have clearer error messaging; now checks for deceased and withdrawn participants and returns with an appropriate error
* storeKitReceipt now checks for deceased and withdrawn participants and returns with an appropriate error; errors are now thrown instead of returned, avoiding an issue with errors being swallowed

### DIFF
--- a/utils/firestore.js
+++ b/utils/firestore.js
@@ -3024,7 +3024,8 @@ const assignKitToParticipant = async (data) => {
     let kitAssignmentResult;
     const { supplyKitId, kitStatus, uniqueKitID, supplyKitTrackingNum, returnKitTrackingNum,
         assigned, collectionRound, collectionDetails, baseline, bioKitMouthwash, bioKitMouthwashBL1, bioKitMouthwashBL2,
-        kitType, mouthwashKit, dateKitRequested, kitLevel, initialKit, replacementKit1, replacementKit2 } = fieldMapping;
+        kitType, mouthwashKit, dateKitRequested, kitLevel, initialKit, replacementKit1, replacementKit2, 
+        withdrawConsent, destroyData, participantDeceased, yes } = fieldMapping;
 
     await db.runTransaction(async (transaction) => {
         // Check the supply kit tracking number and see if it matches the return kit tracking number
@@ -3136,6 +3137,7 @@ const assignKitToParticipant = async (data) => {
             }
         }
 
+        printDocsCount(otherParticipantsWithThisKit, "assignKitToParticipant; collection: possible duplicate kits");
 
         
         const kitData = {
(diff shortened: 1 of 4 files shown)


Gbarra9 (approved):
Looks good

utils/sites.js
```diff
@@ -350,6 +351,25 @@ const updateParticipantData = async (req, res, authObj) => {
             flatUpdateObj[fieldMapping.participationStatus] = fieldMapping.participantMap.dataDestructionRequested;
         }
 
+        // If participant consent is being withdrawn, data destroyed or participant deceased, remove their home MW kit stuff
+        // This is specifically only if the kit status is initialized or address undeliverable
+        // it will otherwise be handled elsewhere
+        if (
+            flatDocData[fieldMapping.withdrawConsent] == fieldMapping.yes || 
+            flatDocData[fieldMapping.participantMap.destroyData] === fieldMapping.yes ||
```
JoeArmani, line 359: This may be the data destruction requested flag (prior to the participant actually signing the data destruction form). Maybe that's what's wanted here, just flagging because there are a few similarly-worded variables in the DD process.

utils/firestore.js
```diff
@@ -3284,7 +3325,8 @@ const processVerifyScannedCode = async (id) => {
 
 const confirmShipmentKit = async (shipmentData) => {
     try {
-        return await db.runTransaction(async transaction => {
+        let toReturn;
+        await db.runTransaction(async transaction => {
```
JoeArmani, line 3329: Nit (jmo, comment only): One way to make the code's intent a bit clearer would be to:
•omit the `toReturn` variable.
• `return await db.runTransaction(......`
• Then, inside the function, just return whatever conditions that branch generates `return {status: 'This participant is deceased; do not ship this kit. Contact the Biospecimen Team.'};`
•Then the final `return toReturn` would be dead code.

JoeArmani (commented):
Looking good. The only comment I'd like to confirm is the one about `flatDocData` in sites.js.

utils/sites.js
amber-emmes: I believe that is the relevant flag, but now that you mention it, we may in fact want to run this only after the participant has signed the data destruction. Do you know what concept ID + value that would be off the top of your head, or should I go digging?

utils/firestore.js
amber-emmes: I recall this not working as expected when I tried it in another location; this is modeled off of a solution that worked.

PR #822, opened 2025-08-05
DHQ Report data streaming

JoeArmani:
Fix: https://github.com/episphere/connect/issues/1329

Issue found in production:
The 'Detailed Analysis' CSV is very large in prod. It's currently well over 100MB after decompression, and it will grow until the date flags are implemented. This was causing the V8 runtime to crash OOM.

Fix: 
  • Replaced full CSV loading with CSV row-by-row streaming throughout.
  • Memory cleanup of ZIP/CSV data after processing for GC access.
 
This allows processing extremely large CSVs in a memory-constrained environment.

Previously, memory was spiking during 'Detailed Analysis' processing and causing a crash.
The memory logs look steady after this streaming update:

Screenshot 2025-08-05 at 3 52 58 PM

### DIFF
--- a/utils/fileProcessing.js
+++ b/utils/fileProcessing.js
@@ -1,3 +1,6 @@
+const { Readable } = require('stream');
+const readline = require('readline');
+
 /**
  * Extracts files from a ZIP archive. Returns an array of file objects.
  * @param {string} base64Data - Base64 encoded ZIP file data.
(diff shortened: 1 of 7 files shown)


utils/fileProcessing.js
```diff
@@ -247,9 +250,78 @@ const validateCSVRow = (row, requiredFields) => {
...
+    const input = typeof csvContent === 'string' ? csvContent : csvContent.toString('utf8');
+
+    // Create the stream
+    const rl = readline.createInterface({
+        input: Readable.from(input),
+        crlfDelay: Infinity,
+    });
+
+    // Process the stream
+    for await (const rawLine of rl) {
```
we-ai, line 273: Should dev team saves each row as is, and let analytics do data QC and cleanups?

JoeArmani: There was just one edge case I needed to handle reading this raw data, where commas are embedded in the actual response string. We're providing that one as-is, I just had to ensure the commas were identified as being within the string so the rows are consistent.
Screenshot 2025-07-23 at 3 10 23 PM

There was another data flattening issue with the 'Detailed Analysis' file that Nicole asked us to pre-flatten, and I handled that by parsing down to the question_foodId level because that's a direct table translation they can work with.

PR #824, opened 2025-08-08
1326 support

amber-emmes:
Backend implementation of Handling participants with Undeliverable Address mouthwash kit status (Biospecimens)
[#1306](https://github.com/episphere/connect/issues/1306)

Changes:

dashboard.js
* Added new retrieveRequestAKitConditions endpoint to retrieve requestAKitConditions documents. As implemented, ID does not need to be passed in if there is only one such document in the app, but does if there are multiple.
* Added new updateRequestAKitConditions endpoint to update requestAKitConditions documents. As implemented, ID does not need to be passed in if there is only one such document in the app, but does if there are multiple.
* Added new processRequestAKitConditions endpoint. Will run the query built in requestAKitConditions; may update the relevant flags on the participants found or not depending on provided updateDb flag. As implemented, ID does not need to be passed in if there is only one such document in the app, but does if there are multiple.

utils/fieldToConceptIdMapping.js:
* Added kitRequestEligible and dtEligibleToKitRequest entries for new concept IDs added as a part of this functionality.

utils/firestore.js:
* Added retrieveRequestAKitConditions function to retrieve requestAKitConditions documents from Firestore. As implemented, ID does not need to be passed in if there is only one such document in the app, but does if there are multiple. Also has the option to select a more limited number of fields to limit data reads if needed.
* Added updateRequestAKitConditions function to update requestAKitConditions documents in Firestore. As implemented, ID does not need to be passed in if there is only one such document in the app, but does if there are multiple.
* Added processRequestAKitConditions function to retrieve requestAKitConditions documents from Firestore and then run their query conditions, plus some hardcoded conditions, against Google Bigquery. Depending on the passed-in flags, may do a dry run or may update the kitRequestEligible and dtEligibleToKitRequest  flags on the participants. Will return the count of participants found and the query generated for end user viewing. Dry runs do not use the limit specified on the requestAKitConditions document per request.

utils/bigquery.js:
* Added getParticipantsForRequestAKitBQ function to take in requestAKitConditions information, process it into a SQL query, run it against Google BigQuery, and then return the participant rows found and the query string that is run.


utils/dashboard.js
```diff
@@ -134,6 +134,52 @@ const dashboard = async (req, res) => {
     } else if (api === 'getSiteNotification' && isHelpDesk === false) { // Everyone except HelpDesk
         const { getSiteNotification } = require('./notifications');
         return await getSiteNotification(req, res, authObj);
+    } else if (api === 'retrieveRequestAKitConditions') {
+        if (req.method === "OPTIONS") return res.status(200).json({ code: 200 });
```
we-ai, line 138: Checked in line 7.

utils/dashboard.js
```diff
@@ -134,6 +134,52 @@ const dashboard = async (req, res) => {
...
+
+        const { retrieveRequestAKitConditions } = require('./firestore');
+        try {
+            const data = await retrieveRequestAKitConditions(req.query?.docId);
+            return res.status(200).json({ data, code: 200 });
+        } catch(error) {
+            return res.status(500).json({ data: {}, message: error.message, code: 500 });
+        }
+    } else if (api === 'updateRequestAKitConditions') {
+        if (req.method === "OPTIONS") return res.status(200).json({ code: 200 });
```
we-ai, line 152: Checked in line 7.

utils/dashboard.js
```diff
@@ -134,6 +134,52 @@ const dashboard = async (req, res) => {
...
+
+        if (req.method !== "POST") return res.status(405).json(getResponseJSON("Only POST requests are accepted!", 405));
+
+        if (!authObj) return res.status(401).json(getResponseJSON("Authorization failed!", 401));
```
we-ai, line 156: `authObj` is truthy.

utils/dashboard.js
```diff
@@ -134,6 +134,52 @@ const dashboard = async (req, res) => {
...
+            return res.status(200).json({ success, docId, code: 200 });
+        } catch(error) {
+            return res.status(500).json({ success: false, message: error.message, code: 500 });
+        }
+    } else if (api === 'processRequestAKitConditions') {
+        if (req.method === "OPTIONS") return res.status(200).json({ code: 200 });
+
+        if (req.method !== "GET") return res.status(405).json(getResponseJSON("Only GET requests are accepted!", 405));
+
+        if (!authObj) return res.status(401).json(getResponseJSON("Authorization failed!", 401));
```
we-ai, line 172: Same as above.

utils/firestore.js
```diff
@@ -2507,6 +2507,121 @@ const getSiteDetailsWithSignInProvider = async (acronym) => {
     return snapshot.docs[0].data();
 }
 
+const retrieveRequestAKitConditions = async (id, selectFields) => {
+    let query = db.collection("requestAKitConditions");
+    // Currently there is only one requestAKitConditions document per app
+    // but this is built to allow for the possibility
+    // of multiple requestAKitConditions being created in the future
+    if(id) {
```
we-ai, line 2515:
```suggestion
    if (id) {
```

utils/firestore.js
```diff
@@ -2507,6 +2507,121 @@ const getSiteDetailsWithSignInProvider = async (acronym) => {
...
+        query = query.doc(id);
+    }
+    if(selectFields) {
+        query = query.select(...selectFields);
+    }
+    const snapshot = await query.get();
```
we-ai, line 2521: When `id` is missing, will this query gets all documents?

utils/firestore.js
```diff
@@ -2507,6 +2507,121 @@ const getSiteDetailsWithSignInProvider = async (acronym) => {
...
+            [`${verificationStatus}`, 'equals', verified], // Participant is verified
+            [`${withdrawConsent}`, 'notequals', yes], // Participant has not withdrawn consent
+            [`${destroyData}`, 'notequals', yes], // Participant data is not being destroyed
+            [`${participantDeceased}`, 'notequals', yes], // Participant is not deceased
+            [`${participantDeceasedNORC}`, 'notequals', yes], // Participant is not deceased per NORC
+            [`${activityParticipantRefusal}.${baselineMouthwashSample}`, 'notequals', yes], // Participant has not refused baseline mouthwash sample
+            [`${activityParticipantRefusal}.${allFutureSamples}`, 'notequals', yes],// Participant has not refused all future samples
+            [`${refusedAllFutureActivities}`, 'notequals', yes],// Participant has not refused all future activities
+            `NOT REGEXP_CONTAINS(d_${physicalAddress1}, ${JSON.stringify(poBoxRegex.toString())}) OR NOT REGEXP_CONTAINS(d_${address1}, ${JSON.stringify(poBoxRegex.toString())})`, // Participant address is not a P.O. Box
+            `d_${isPOBox}.integer != ${yes}`,// PO Box is not checked
```
we-ai, line 2565: `d_CID.integer` will break after this (mixed data types) is fixed.

utils/firestore.js
amber-emmes: It will pull all documents, and if more than one is found, throw an error informing the user to specify an ID.

we-ai: > It will pull all documents, and if more than one is found, throw an error informing the user to specify an ID.

Not sure how often there'll be only one or 0 document in collection. Should we check document count first in this scenario?

utils/firestore.js
we-ai: > I'm not sure.I follow, sorry. What change should I make?

I meant `d_${isPOBox}.integer != ${yes}`. It should be `d_${isPOBox} != ${yes}` after data fix.
And we should fix mixed data types like this.

amber-emmes: Who would I talk to about fixing these data types and confirming that this is not an issue in stage and prod?

utils/firestore.js
amber-emmes: Under the current setup, it should not be possible to create more than 1 of these documents in an environment unless it is manually created through the Firestore UI.

utils/firestore.js
amber-emmes: Code updated.

PR #845, opened 2025-09-30
Updated retrieveNotifications

bransteitterbr:
The retrieveNotifications will now only update the read status if requested via a query parameter


amber-emmes (approved):
Looks good.

PR #926, opened 2026-03-27
Limit DOB Check

### LINKED ISSUE #1604: Prod issue: identifyParticipant API blocking records due to DOB requirement
As an unintended consequence of blocking out of range or missing DOB in API for identifyParticipant API in the March release, we are inadvertently blocking verification statuses other than 'verified' for records that have out of range or missing DOB.

The API needs to allow missing or out of range DOB for verification statuses other than Verified.

bransteitterbr:
For Issue 1604: https://github.com/episphere/connect/issues/1604

Only verifying the DOB if the new status is verified


amber-emmes (approved):
Looks good.

JoeArmani (approved):
Looks good

PR #951, opened 2026-05-28
Dev -> Stage site notifications edge case & tests

JoeArmani:
Dev -> Stage site notifications edge case & tests

Handle matching to & cc/bcc edge case:
•Handle the dev case where to and cc/bcc match (dedupe them).
•Don't allow a failed notification to cause a failed Firestore update.
•Improve error reporting on failed email notifications.

Update Cloud Tasks handling
Add TTL using expiresAt with 45 day cleanup to notificationBulkRuns and their batches subcollections (the documents that track bulk notification runs through their lifecycle).

### DIFF
--- a/test/mocks/helpers/mockHelpers.js
+++ b/test/mocks/helpers/mockHelpers.js
@@ -60,6 +60,29 @@ class MockHelpers {
                 },
                 Transaction: function Transaction() {},
                 Filter: {},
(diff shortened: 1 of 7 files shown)


amber-emmes (approved):
Looks good.

we-ai (approved):

PR #973, opened 2026-07-16
Quick Casing Fix

amber-emmes:
Fixed an issue where the code was using fieldMapping.healthcareProvider instead of fieldMapping.healthCareProvider, resulting in the value not being found.


bransteitterbr (approved):

JoeArmani (approved):
Looks good

PR #977, opened 2026-07-17
Diet Screener should be set to notStarted at the time of verification

bransteitterbr:
For Issue 1550: https://github.com/episphere/connect/issues/1550
Per the spec the Diet Screener survey should be marked as notStarted at the time the user is verified


Gbarra9 (approved):
looks good

we-ai (approved):

